PR #26, opened 2023-01-05
MockResponseProvider

matejmolnar:
- add ResponseProviding protocol, a type that is able to provide a `Response` for a given `URLRequest`.
- add MockResponseProvider which creates responses for requests from corresponding data files stored in Assets (for testing purposes).
- conform URLSession to ResponseProviding and use it as default in APIManager.

### DIFF
--- a/NetworkingSampleApp/NetworkingSampleApp/Scene/SampleViewScene/SampleViewModel.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scene/SampleViewScene/SampleViewModel.swift
@@ -23,7 +23,7 @@ final class SampleViewModel {
         #endif
         
         return APIManager(
-            urlSession: URLSession.shared,
+            responseProvider: URLSession.shared,
             requestAdapters: [loggingInterceptor],
             responseProcessors: responseProcessors,
             errorProcessors: errorProcessors
--- a/NetworkingSampleApp/NetworkingSampleApp/Scene/SampleViewScene/SampleViewModel.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scene/SampleViewScene/SampleViewModel.swift
@@ -23,7 +23,7 @@ final class SampleViewModel {
         #endif
         
         return APIManager(
-            responseProvider: URLSession.shared,
+            urlSession: URLSession.shared,
             requestAdapters: [loggingInterceptor],
             responseProcessors: responseProcessors,
             errorProcessors: errorProcessors
--- a/Sources/Networking/Core/APIManager.swift
+++ b/Sources/Networking/Core/APIManager.swift
@@ -12,19 +12,19 @@ open class APIManager: APIManaging {
     private let requestAdapters: [RequestAdapting]
     private let responseProcessors: [ResponseProcessing]
     private let errorProcessors: [ErrorProcessing]
-    private let urlSession: URLSession
+    private let responseProvider: ResponseProviding
     private let sessionId: String
     private var retryCounter = Counter()
     
     public init(
-        urlSession: URLSession = URLSession(configuration: .default),
+        responseProvider: ResponseProviding = URLSession(configuration: .default),
         requestAdapters: [RequestAdapting] = [],
         responseProcessors: [ResponseProcessing] = [StatusCodeProcessor()],
         errorProcessors: [ErrorProcessing] = []
     ) {
         /// generate session id in readable format
         sessionId = Date().ISO8601Format()
-        self.urlSession = urlSession
+        self.responseProvider = responseProvider
         self.requestAdapters = requestAdapters
         self.responseProcessors = responseProcessors
         self.errorProcessors = errorProcessors
@@ -48,7 +48,7 @@ private extension APIManager {
             request = try await requestAdapters.adapt(request, for: endpointRequest)
             
             /// call request on url session
-            var response = try await urlSession.data(for: request)
+            var response = try await responseProvider.response(for: request)
             
             /// process request
             response = try await responseProcessors.process(response, with: request, for: endpointRequest)
--- a/Sources/Networking/Core/APIManager.swift
+++ b/Sources/Networking/Core/APIManager.swift
@@ -47,7 +47,7 @@ private extension APIManager {
             /// adapt request with all adapters
             request = try await requestAdapters.adapt(request, for: endpointRequest)
             
-            /// call request on url session
+            /// get response for given request (usually fires a network request via URLSession)
             var response = try await responseProvider.response(for: request)
             
             /// process request
--- a/Sources/Networking/Core/APIManager.swift
+++ b/Sources/Networking/Core/APIManager.swift
@@ -17,7 +17,21 @@ open class APIManager: APIManaging {
     private var retryCounter = Counter()
     
     public init(
-        responseProvider: ResponseProviding = URLSession(configuration: .default),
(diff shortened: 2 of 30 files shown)


Tests/NetworkingTests/SampleResponseProviderTests.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+//  Created by Matej Molnár on 05.01.2023.
+//
+
+@testable import Networking
+import XCTest
+
+final class SampleResponseProviderTests: XCTestCase {
+    // swiftlint:disable:next force_unwrapping
+    private lazy var mockUrlRequest = URLRequest(url: URL(string: "https://reqres.in/api/users?page=2")!)
```
cejanen, line 13: besides this I would like also have integration test which will cover the flow  ( mock data -> store it via StorageProcessor -> load via this Response provider and test that final data are the same as at the beginning. 
Right now we have two different objects which stores/read data from file system and would like to avoid any discrepancy in future

cejanen (changes requested):
Nice, just asking for one more test

Tests/NetworkingTests/SampleResponseProviderTests.swift
matejmolnar: I agree, but for this test flow we would need a new `ResponseProvider` because the `SampleResponseProvider` only loads data from Assets but `EndpointRequestStorageProcessor` stores data into file system.

cejanen (approved):
After some discussion it's fine.
 
 🥐

Sources/Networking/Core/APIManager.swift
```diff
@@ -12,19 +12,19 @@ open class APIManager: APIManaging {
...
     private let responseProcessors: [ResponseProcessing]
     private let errorProcessors: [ErrorProcessing]
-    private let urlSession: URLSession
+    private let responseProvider: ResponseProviding
     private let sessionId: String
     private var retryCounter = Counter()
     
     public init(
-        urlSession: URLSession = URLSession(configuration: .default),
+        responseProvider: ResponseProviding = URLSession(configuration: .default),
```
schwarja, line 20: I suggest having 2 initializers. One with the old signature with `urlSession: URLSession` and the new one. They would be almost duplicate but I find the `responseProvider` parameter name extremely misleading if you want to initialize an API manager just for standard use with a URL session

Sources/Networking/Core/SampleResponseProvider.swift
```diff
@@ -0,0 +1,80 @@
...
+#if os(macOS)
+    import AppKit
+#else
+    import UIKit
+#endif
+
+// MARK: - SampleResponseProvider definition
+
+/// A response provider which creates responses for requests from corresponding data files stored in Assets.
+public class SampleResponseProvider: ResponseProviding {
```
schwarja, line 20: Let's make it `open` again, plus, can we rename it to `DefaultResponseProvider` or just `ResponseProvider`? I don't believe there should be anything called `sample` in the package itself

Sources/Networking/Misc/SampleResponseProviderError.swift
```diff
@@ -0,0 +1,16 @@
...
+//  SampleResponseProviderError.swift
+//  
+//
+//  Created by Matej Molnár on 04.01.2023.
+//
+
+import Foundation
+
+/// An error that occurs during loading a ``Response`` from assets by `SampleResponseProvider`.
+enum SampleResponseProviderError: Error {
```
schwarja, line 11: Please get rid of the `sample` prefix also here

NetworkingSampleApp/NetworkingSampleApp/Scene/SampleViewScene/SampleViewModel.swift
```diff
@@ -23,7 +23,7 @@ final class SampleViewModel {
         #endif
         
         return APIManager(
-            urlSession: URLSession.shared,
+            responseProvider: URLSession.shared,
```
schwarja, line 26: As mentioned in `APIManager` I would rather see the `urlSession` parameter name here

Tests/NetworkingTests/ErrorProcessorTests.swift
```diff
@@ -113,7 +113,7 @@ final class ErrorProcessorTests: XCTestCase {
 extension ErrorProcessorTests {
     func test_apiManager_request_errorShouldBeMappedToSimpleError() async {
         let apiManager = APIManager(
-            urlSession: URLSession.shared,
+            responseProvider: URLSession.shared,
```
schwarja, line 116: As mentioned in `APIManager` I would rather see the `urlSession` parameter name here

Tests/NetworkingTests/ErrorProcessorTests.swift
```diff
@@ -129,7 +129,7 @@ extension ErrorProcessorTests {
     
     func test_apiManager_request_originalErrorShouldRemain() async {
         let apiManager = APIManager(
-            urlSession: URLSession.shared,
+            responseProvider: URLSession.shared,
```
schwarja, line 132: As mentioned in `APIManager` I would rather see the `urlSession` parameter name here

Tests/NetworkingTests/SampleResponseProviderTests.swift
```diff
@@ -0,0 +1,72 @@
...
+//  SampleResponseProviderTests.swift
+//  
+//
+//  Created by Matej Molnár on 05.01.2023.
+//
+
+@testable import Networking
+import XCTest
+
+final class SampleResponseProviderTests: XCTestCase {
```
schwarja, line 11: Get rid of the `sample` prefix

Tests/NetworkingTests/SampleResponseProviderTests.swift
```diff
@@ -0,0 +1,72 @@
...
+        "cf-cache-status" : "HIT",
+        "Cache-Control" : "max-age=14400",
+        "Access-Control-Allow-Origin" : "*",
+        "cf-ray" : "784545f34d2f27bc-PRG",
+        "Date" : "Wed, 04 Jan 2023 16:15:29 GMT",
+        "Via" : "1.1 vegur",
+        "Age" : "6306"
+    ]
+    
+    func testLoadingData() async throws {
```
schwarja, line 33: Could we also test some failure? 🙏

schwarja (changes requested):
Looks great 🤩 Just a few minor comments

Sources/Networking/Core/SampleResponseProvider.swift
cejanen: About accessory levels I left some note. We may unify our approach overall. The rest I agree

matejmolnar: @schwarja @cejanen I think that `DefaultResponseProvider` is misleading since the default ResponseProvider is a `URLSession` instance. `ResponseProvider` is also not a great name because we may have multiple implementations of `ResponseProviding`. Maybe `AssetResponseProvider` or `AssetLoadingResponseProvider` would fit better. What do you think?

schwarja: `MockResponseProvider`?

matejmolnar: Makse sense 👍

schwarja (approved):
🦉

PR #28, opened 2023-01-16
Download support

matejmolnar:


Sources/Networking/Core/APIManager.swift
```diff
@@ -50,17 +50,119 @@ open class APIManager: APIManaging {
...
+            }
+            
+            downloadObserver.errorHandler = { [errorProcessors] error in
+                if let resumableData = (error as? URLError)?.userInfo[NSURLSessionDownloadTaskResumeData] as? Data {
+                    continuation.yield(.terminated(resumableData: resumableData))
+                }
+                
+                Task {
+                    /// error processing
+                    let error = await errorProcessors.process(error, for: endpointRequest)
```
matejmolnar, line 146: I've added a retry logic in case something fails during the request creating or response processing, however I am not convinced that we should also add retry logic here in case the download fails while already in progress. Handling temporary loss of network while on poor connection should be handled automatically by URLSession based on it's configuration by setting an appropriate timeout interval for example. For downloads of larger files a background URLSessionConfiguration should be used https://developer.apple.com/documentation/foundation/url_loading_system/downloading_files_in_the_background
What do you think @cejanen?

Sources/Networking/Misc/DownloadState.swift
```diff
@@ -0,0 +1,14 @@
...
+//  
+//
+//  Created by Matej Molnár on 16.01.2023.
+//
+
+import Foundation
+
+public enum DownloadState {
+    case progress(downloadedBytes: Double, totalBytes: Double)
+    case completed(Data?)
```
cejanen, line 12: i feel a bit confusion in states completed and terminated... when completed - that means finished with success we have Data as optional - so not sure if there are any... but for terminated - which I understand is fail we have Data type - so when we fail we always have data (to be able to continue) but when we finish we are not sure if we have data...

Sources/Networking/Utils/DownloadObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+
+import Foundation
+
+enum DownloadObserverError: Error {
+    case missingResponse
+}
+
+/// A helper object which wraps URLSessionDownloadDelegate method callbacks into completion blocks/async functions.
+open class DownloadObserver: NSObject {
```
cejanen, line 15: in this file I would like to add some MARKs to have nicer file structure etc

Sources/Networking/Utils/DownloadObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+    public var progressHandler: ((_ totalBytesWritten: Int64, _ totalBytesExpectedToWrite: Int64) -> ())?
+    public var errorHandler: ((Error) -> ())?
+    public var completionHandler: ((Data?) -> ())?
+    
+    private var responseHandler: ((URLResponse?, Error?) -> ())?
+}
+
+extension DownloadObserver {
+    func response() async throws -> URLResponse {
+        return try await withCheckedThrowingContinuation({ continuation in
```
cejanen, line 25: do we need this return?

Sources/Networking/Utils/DownloadObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+                    return
+                }
+                
+                continuation.resume(throwing: DownloadObserverError.missingResponse)
+            }
+        })
+    }
+}
+
+extension DownloadObserver: URLSessionDownloadDelegate {
```
cejanen, line 43: why are methods in this extension scope with public accessory type?

cejanen (changes requested):
Good attempt. I believe we are on good way even though the fact there is no way to use background queue and URLTask delegate aka
https://www.ralfebert.com/ios-examples/networking/urlsession-background-downloads/

But I think that overall approach is fine - special methods for download on response provider(URLSession), we can use standard adapters approach for request and we abstract a bit to be able to use processor and error handling with retry for first response (response header) before data downloading itself

PR #33, opened 2023-03-07
Download support

matejmolnar:
A rough implementation of a working `DownloadAPIManager` with an example scene.

- Separate manager to deal with download tasks DownloadAPIManager
- It's now possible to use background urlSession configuration specific for downloading files
- Retry handling moved to a protocol Retryable to avoid duplicate code
- Download simulation UI to test download tasks with the option to run/suspend/cancel


Sources/Networking/Misc/DownloadState.swift
```diff
@@ -0,0 +1,37 @@
...
+//  
+//
+//  Created by Matej Molnár on 07.03.2023.
+//
+
+import Foundation
+
+public extension URLSessionTask {
+    struct DownloadState {
+        public var totalBytesWritten: Int64
```
cejanen, line 12: I would vote for better naming here, something like totalBytes vs downloadedBytes?

Sources/Networking/Misc/DownloadState.swift
```diff
@@ -0,0 +1,37 @@
+//
+//  DownloadState.swift
```
cejanen, line 2: The file name should tell it's URLSessionTask extension

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadsViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+//  
+//
+//  Created by Matej Molnár on 07.03.2023.
+//
+
+import Foundation
+import Networking
+import OSLog
+
+extension DownloadAPIManager {
```
cejanen, line 12: I don't see any additional value having it in separated extension in this class, for sample we can init the DownloadManager in DownloadsViewModel

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+}
+
+// MARK: Public API
+public extension DownloadAPIManager {
+    func downloadRequest(
+        _ endpoint: Requestable,
+        resumableData: Data? = nil,
+        retryConfiguration: RetryConfiguration?
+    ) async throws -> (URLSessionDownloadTask, Response) {
```
cejanen, line 62: Do you think may be worth to typealias return tuple into something easier to read? Similar we have Response = (Data, URLResponse)?

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -0,0 +1,220 @@
...
+//  
+//
+//  Created by Matej Molnár on 07.03.2023.
+//
+
+import Foundation
+import Combine
+
+/// Default Download API manager
+open class DownloadAPIManager: NSObject, Retryable {
```
cejanen, line 12: I would add a little bit more description here, how it works, not necessary proper documentation, that will be revalidated as separate issue

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -0,0 +1,220 @@
...
+        _ endpoint: Requestable,
+        resumableData: Data? = nil,
+        retryConfiguration: RetryConfiguration?
+    ) async throws -> (URLSessionDownloadTask, Response) {
+        /// create identifiable request from endpoint
+        let endpointRequest = EndpointRequest(endpoint, sessionId: sessionId)
+        return try await downloadRequest(endpointRequest, resumableData: resumableData, retryConfiguration: retryConfiguration)
+    }
+    
+    func progressStream(for task: URLSessionTask) -> AsyncStream<URLSessionTask.DownloadState> {
```
cejanen, line 68: I haven't find a place in which is this method used? Is it a dead code? (just checking PR via github)

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    public func urlSession(_: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        downloadStateDict[task]?.error = error
+        updateTasks()
+    }
+}
+
+// MARK: URLSessionTask + asyncResponse
+private extension URLSessionTask {
```
cejanen, line 191: why we haven't put this wrapper into URLSessionTask extension we have and we keep it private here?

cejanen (changes requested):
Hola, 👏 
thank you, finally checked the PR, I would say the current approach is feasible and nice. I found minor changes we can improve

NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleDownloadRouter.swift
```diff
@@ -0,0 +1,42 @@
...
+import Networking
+
+/// Implementation of sample API router
+enum SampleDownloadRouter: Requestable {
+    case download(url: URL)
+    
+    var baseURL: URL {
+        switch self {
+        case let .download(url):
+            return url
```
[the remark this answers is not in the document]
gajddo00: which new version?

Sources/Networking/Core/Requestable+Convenience.swift
```diff
@@ -48,7 +48,14 @@ public extension Requestable {
 public extension Requestable {
     func urlComponents() throws -> URLComponents {
         // url creation
-        let urlPath = baseURL.appendingPathComponent(path)
+        let urlPath = {
```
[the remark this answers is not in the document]
gajddo00: This is more or less unusual and edge case situation, so I wouldn't make it convenience. I would just simplify the statement to use a ternary operator.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadsViewModel.swift
gajddo00: The download manager instance is used in multiple places. I agree that it does not make sense having this extension in the DownloadsViewModel so I moved it to a separate file.

Sources/Networking/Core/DownloadAPIManager.swift
gajddo00: Yeah actually realised the protocol for this is still missing, I'll add that and add the description there.

Sources/Networking/Core/DownloadAPIManager.swift
gajddo00: [Here](https://github.com/strvcom/ios-networking/blob/880aec15e7a23c564b3af188d77deebefd293f99/NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressViewModel.swift#L22)

Sources/Networking/Core/DownloadAPIManager.swift
gajddo00: agreed, moved and kept internal

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressViewModel.swift
```diff
@@ -0,0 +1,75 @@
+//
+//  DownloadRowViewModel.swift
+//  NetworkingSampleApp
+//
+//  Created by Matej Molnár on 07.03.2023.
+//
+
+import SwiftUI
```
tonyskansf, line 8:
```suggestion
import Foundation
```

Btw. is it becoming a thing to import UI framework into the view model nowadays?

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressViewModel.swift
```diff
@@ -0,0 +1,75 @@
+//
+//  DownloadRowViewModel.swift
```
tonyskansf, line 2:
```suggestion
//  DownloadProgressViewModel.swift
```

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressViewModel.swift
```diff
@@ -0,0 +1,75 @@
...
+    private let task: URLSessionTask
+    
+    @Published var state: DownloadProgressState = .init()
+    
+    init(task: URLSessionTask) {
+        self.task = task
+    }
+    
+    func onAppear() {
+        Task {
```
brenovaladao, line 21: you might be able to make this function async and call it on the `.task` instead of onAppear

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressViewModel.swift
```diff
@@ -0,0 +1,75 @@
...
+            let stream = DownloadAPIManager.shared.progressStream(for: task)
+
+            for try await downloadState in stream {
+                DispatchQueue.main.async { [weak self] in
```
brenovaladao, line 25: can't this be done in a main actor task?

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressView.swift
```diff
@@ -0,0 +1,62 @@
...
+        .padding(10)
+        .background(
+            Color.white
+                .cornerRadius(15)
+                .shadow(radius: 10)
+        )
+        .padding(15)
+        .onAppear {
+            viewModel.onAppear()
+        }
```
brenovaladao, line 60:
```suggestion
        .task {
            await viewModel.onAppear()
        }
```

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadsViewModel.swift
```diff
@@ -0,0 +1,41 @@
...
+import OSLog
+
+@MainActor
+final class DownloadsViewModel: ObservableObject {
+    @Published var tasks: [URLSessionTask] = []
+    @Published var urlText: String = "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ElephantsDream.mp4"
+    private let downloadAPIManager = DownloadAPIManager.shared
+    
+    func onAppear() {
+        Task {
```
brenovaladao, line 19: same here

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -0,0 +1,193 @@
...
+                .sink(receiveValue: { dict in
+                    guard let downloadState = dict[task] else {
+                        return
+                    }
+                    
+                    continuation.yield(downloadState)
+                    
+                    if downloadState.error != nil ||
+                        downloadState.downloadedFileURL != nil
+                    {
```
brenovaladao, line 82: this feels like a mix of indentation styles 🤔

brenovaladao (commented):
🪙

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -0,0 +1,193 @@
...
+        }
+    }
+}
+
+// MARK: URLSession Delegate
+extension DownloadAPIManager: URLSessionDelegate, URLSessionDownloadDelegate {
+    public func urlSession(_: URLSession, downloadTask: URLSessionDownloadTask, didWriteData _: Int64, totalBytesWritten: Int64, totalBytesExpectedToWrite: Int64) {
+        downloadStateDict[downloadTask]?.downloadedBytes = totalBytesWritten
+        downloadStateDict[downloadTask]?.totalBytes = totalBytesExpectedToWrite
+    }
```
tonyskansf, line 182: Could this be the place  where (accessing the `downloadStateDict`) it's not thread-safe and perhaps causing the crash? Maybe using an actor would be safer.

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -0,0 +1,193 @@
...
+import Combine
+
+/// Default Download API manager
+open class DownloadAPIManager: NSObject, Retryable {
+    private let requestAdapters: [RequestAdapting]
+    private let responseProcessors: [ResponseProcessing]
+    private let errorProcessors: [ErrorProcessing]
+    private var urlSession: URLSession!
+    private let sessionId: String
+    private var taskStateCancellables: [URLSessionTask: AnyCancellable] = [:]
```
tonyskansf, line 18: Suggestion: I'd keep lets and vars together. 😄

Sources/Networking/Misc/URLSessionTask+asyncResponse.swift
```diff
@@ -0,0 +1,40 @@
...
+                        if let response {
+                            continuation.resume(returning: response)
+                        }
+                    }
+                }
+            },
+            onCancel: { [cancellable] in
+                cancellable?.cancel()
+            })
+    }
```
tonyskansf, line 39:
```suggestion
            onCancel: { [cancellable] in
                cancellable?.cancel()
            }
        )
    }
```
Indentation

Sources/Networking/Misc/URLSessionTask+asyncResponse.swift
```diff
@@ -0,0 +1,40 @@
...
+                        response != nil || error != nil
+                    })
+                    .sink { (response, error) in
+                        if let error {
+                            continuation.resume(throwing: error)
+                        }
+                        
+                        if let response {
+                            continuation.resume(returning: response)
+                        }
```
tonyskansf, line 32:
```suggestion
                       if let error {
                            continuation.resume(throwing: error)
                       } else if let response {
                            continuation.resume(returning: response)
                       }
```

Continuations should be resumed only once otherwise it can corrupt data. Although both probably won't be non-nil at the same time, I'd suggest writing an if-else condition to be safe.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressView.swift
```diff
@@ -0,0 +1,62 @@
+//
+//  DownloadRow.swift
```
tonyskansf, line 2:
```suggestion
//  DownloadProgressView.swift
```

tonyskansf (changes requested):
👋 

The sample app crashes on both the simulator and real device when multiple downloads are ongoing (at 5 downloads). Reproducible on every try. 

See [crash logs](https://gist.github.com/tonyskansf/8cc23933112ea228cb8d34e0f848e9fe).

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressView.swift
cejanen: 👍 thinking also about better name for viewModel function, maybe startDownloadTasks maybe better? Or any other suggestions?

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressViewModel.swift
cejanen: Nice catch, please no UI related stuff on VMs, maybe some exceptions in UIKit(eg I saw VM as diffableDataSource provider and that we can overcome 💁‍♂️)

Sources/Networking/Core/DownloadAPIManager.swift
cejanen: 👍 similar we have for running calls and retry mechanism

Sources/Networking/Misc/URLSessionTask+asyncResponse.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  File.swift
```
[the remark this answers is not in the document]
cejanen: I would rename file with uppercase "a" as URLSessionTask+AsyncResponse.swift to keep extensions file naming consistent

cejanen (changes requested):
My pieces also added 
Thanks guys for great comments

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressView.swift
gajddo00: It's not starting the download, so I'd suggest something like observeDownloadProgress or startObservingDownloadTask.

Sources/Networking/Core/DownloadAPIManager.swift
gajddo00: agreed, all props which are being mutated should be thread safe

tonyskansf (commented):
The crash is now fixed with the actors 👍

From further testing, the memory graph showed that there is a memory leak — the `DownloadAPIManager` is not being released.

image

This is what I've changed to test it:
- Initialize the `DownloadAPIManager` only once in the `DownloadsViewModel` and pass this instance down to the `DownloadProgressViewModel` instead of using `shared`.

Going from the examples list to Downloads and back (without tapping "Download") will cause the memory leak. This seems to be the source:

```swift
urlSession = URLSession(
    configuration: urlSessionConfiguration,
    delegate: self,
    delegateQueue: OperationQueue()
)
```

gajddo00:
By design, the DownloadAPIManager should be used as a singleton object. URLSession keeps strong reference and Apple also recommends using a shared urlSession, otherwise we'll need the user to free the urlSession manually by calling either
- urlSession.finishTasksAndInvalidate()
- invalidateAndCancel()

I added this as a option in case somebody would like to use a non-singleton instance in the future so they could invalidate the session manually by calling
```
deinit {
      downloadAPIManager.invalidateSession(true)
}
```

Give it a try 👍

brenovaladao:
When testing the app on the simulator, when I tap on the `download` button it always starts the download twice, is it the intended behavior

I'm 100% sure I'm tapping just once, tried several times


https://github.com/strvcom/ios-networking/assets/46816629/73f36457-11da-47fc-a65c-4228dced7fdc


The logs

```
2023-05-30 12:54:09.589072+0200 NetworkingSampleApp[37700:456392] 🔽🔽🔽 REQUEST  🔽🔽🔽
2023-05-30 12:54:09.589072+0200 NetworkingSampleApp[37700:456021] 🔽🔽🔽 REQUEST  🔽🔽🔽
2023-05-30 12:54:09.589267+0200 NetworkingSampleApp[37700:456392] 🔈 GET http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ElephantsDream.mp4
2023-05-30 12:54:09.589274+0200 NetworkingSampleApp[37700:456021] 🔈 GET http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ElephantsDream.mp4
2023-05-30 12:54:09.589397+0200 NetworkingSampleApp[37700:456021] 🔼🔼🔼 REQUEST END 🔼🔼🔼
2023-05-30 12:54:09.589397+0200 NetworkingSampleApp[37700:456392] 🔼🔼🔼 REQUEST END 🔼🔼🔼
2023-05-30 12:54:09.929058+0200 NetworkingSampleApp[37700:456394] ✅✅✅ RESPONSE ✅✅✅
2023-05-30 12:54:09.929141+0200 NetworkingSampleApp[37700:456394] 🔈 200 GET http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ElephantsDream.mp4
2023-05-30 12:54:09.929323+0200 NetworkingSampleApp[37700:456394] 👉 Date: Tue, 30 May 2023 10:54:09 GMT
2023-05-30 12:54:09.929427+0200 NetworkingSampleApp[37700:456394] 👉 Server: UploadServer
2023-05-30 12:54:09.929528+0200 NetworkingSampleApp[37700:456394] 👉 Content-Type: video/mp4
2023-05-30 12:54:09.929622+0200 NetworkingSampleApp[37700:456394] 👉 Content-Length: 169612362
2023-05-30 12:54:09.929758+0200 NetworkingSampleApp[37700:456394] ✅✅✅ RESPONSE END ✅✅✅
2023-05-30 12:54:09.947203+0200 NetworkingSampleApp[37700:456394] 🎈 Response saved 1107 bytes at /Users/brenovaladao/Library/Developer/CoreSimulator/Devices/D3208BF0-13D9-4B47-A305-F042C5933E36/data/Containers/Data/Application/75207FE4-5E20-4BA8-86C7-6B905D337CB8/tmp/responses/2023-05-30T10:48:10Z/2023-05-30T10:48:10Z_commondatastorage.googleapis.com_gtv-videos-bucket_sample_elephantsdream.mp4_get_8.json
2023-05-30 12:54:10.161713+0200 NetworkingSampleApp[37700:456393] ✅✅✅ RESPONSE ✅✅✅
2023-05-30 12:54:10.161791+0200 NetworkingSampleApp[37700:456393] 🔈 200 GET http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ElephantsDream.mp4
2023-05-30 12:54:10.161897+0200 NetworkingSampleApp[37700:456393] 👉 Date: Tue, 30 May 2023 10:54:09 GMT
2023-05-30 12:54:10.162070+0200 NetworkingSampleApp[37700:456393] 👉 Server: UploadServer
2023-05-30 12:54:10.162213+0200 NetworkingSampleApp[37700:456393] 👉 Content-Type: video/mp4
2023-05-30 12:54:10.162364+0200 NetworkingSampleApp[37700:456393] 👉 Content-Length: 169612362
2023-05-30 12:54:10.162547+0200 NetworkingSampleApp[37700:456393] ✅✅✅ RESPONSE END ✅✅✅
2023-05-30 12:54:10.175162+0200 NetworkingSampleApp[37700:456395] 🎈 Response saved 1109 bytes at /Users/brenovaladao/Library/Developer/CoreSimulator/Devices/D3208BF0-13D9-4B47-A305-F042C5933E36/data/Containers/Data/Application/75207FE4-5E20-4BA8-86C7-6B905D337CB8/tmp/responses/2023-05-30T10:48:10Z/2023-05-30T10:48:10Z_commondatastorage.googleapis.com_gtv-videos-bucket_sample_elephantsdream.mp4_get_9.json
```

gajddo00:
Yes, that was just exemplary to test out the support for simultaneous downloads. Let's keep it as the original - one download showcase.

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -0,0 +1,219 @@
...
+        _ endpoint: Requestable,
+        resumableData: Data? = nil,
+        retryConfiguration: RetryConfiguration?
+    ) async throws -> DownloadResult {
+        /// create identifiable request from endpoint
+        let endpointRequest = EndpointRequest(endpoint, sessionId: sessionId)
+        return try await downloadRequest(endpointRequest, resumableData: resumableData, retryConfiguration: retryConfiguration)
+    }
+    
+    /// Creates an async stream of download state updates for a given task. ]
```
cejanen, line 78: extra bracket in the comment

cejanen (changes requested):
Another round of minor updates

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadsViewModel.swift
```diff
@@ -0,0 +1,47 @@
...
+import OSLog
+
+@MainActor
+final class DownloadsViewModel: ObservableObject {
+    @Published var tasks: [URLSessionTask] = []
+    @Published var urlText: String = "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ElephantsDream.mp4"
+    private let downloadAPIManager = DownloadAPIManager.shared
+
+    func startDownload() {
+        Task {
```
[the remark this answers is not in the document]
gajddo00: That's already removed, it was to showcase support for simultaneous downloads.

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -0,0 +1,214 @@
...
+                
+                return try await downloadRequest(
+                    endpointRequest,
+                    resumableData: resumableData,
+                    retryConfiguration: retryConfiguration
+                )
+            } catch {
+                /// error processing
+                throw await errorProcessors.process(error, for: endpointRequest)
+            }
```
brenovaladao, line 152: maybe this could go in another method

brenovaladao (approved):
👍🏼

cejanen (approved):
I happy with final result. No code is perfect but we enhanced it a lot in PR. @tonyskansf can you just confirm your concerns about strong reference in URLSession delegate are fixed or explained?
Thank you all!

👏

Sources/Networking/Core/DownloadAPIManaging.swift
```diff
@@ -0,0 +1,53 @@
...
+/// Recommended to be used as singleton.
+/// If you wish to use multiple instances, make sure you manually invalidate url session by calling the `invalidateSession` method.
+public protocol DownloadAPIManaging {
+    /// List of all currently ongoing download tasks.
+    var allTasks: [URLSessionDownloadTask] { get async }
+    
+    /// Invalidates urlSession instance.
+    /// - Parameters:
+    ///   - shouldFinishTasks: Indicates whether all currently active tasks should be able to finish before invalidating. Otherwise they will be cancelled.
+    func invalidateSession(_ shouldFinishTasks: Bool)
```
tonyskansf, line 23:
```suggestion
    func invalidateSession(shouldFinishTasks: Bool)
```

Otherwise `invalidateSession(false)` could be misleading.

tonyskansf (commented):
> By design, the DownloadAPIManager should be used as a singleton object. URLSession keeps strong reference and Apple also recommends using a shared urlSession, otherwise we'll need the user to free the urlSession manually

What specific part of the implementation suggests using the `DownloadAPIManager` as a singleton prior to 4f25236? I thought it was just for the sake of the example app 😄 

Out of curiosity and scope of this PR, is it possible to let the concrete implementation of the `DownloadAPIManaging` handle the URL session instead of leaving this task up to the dev who is error-prone? For example, the URL session could be invalidated upon the deinitialization of the manager instance. That way, the lifecycle of the URL session would be bound to it. 🤔 

```swift
// DownloadAPIManager
deinit {
    invalidateSession()
}

tonyskansf (approved):
> I guess the only way would be to create the session every time a download is requested and then destroy it after the download is finished.

If that would make sense that could be an improvement. I think for now, whoever will use this lib and want this behavior can provide a different `DownloadAPIManaging` implementation or subclass the one provided.

PR #38, opened 2023-05-08
Support for associated array query parameters

gajddo00:
In order to support associated array query parameters (there are no official formal definitions so let's call it that), an ArrayParameter and ArrayEncoding options are introduced.

**ArrayParameter**
- new type introduced to add array encoding support
```
var urlParameters: [String: Any]? { 
     ["filter": ArrayParameter([1, 2, 3], arrayEncoding: .individual)]
}
```

**ArrayEncoding**
- commaSeparated  = filter=1,2,3
- individual = filter=1&filter=2&filter=3
sometimes referred to as OR AND modes.


cejanen:
> In order to support associated array query parameters (there are no official formal definitions so let's call it that), an ArrayParameter and ArrayEncoding options are introduced.
> 
> **ArrayParameter**
> 
> * new type introduced to add array encoding support
> 
> ```
> var urlParameters: [String: Any]? { 
>      ["filter": ArrayParameter([1, 2, 3], arrayEncoding: .individual)]
> }
> ```
> 
> **ArrayEncoding**
> 
> * commaSeparated  = filter=1,2,3
> * individual = filter=1&filter=2&filter=3
>   sometimes referred to as OR AND modes.

I tried to find a name for this approach and asked chatgpt 😆 
The syntax you're referring to, where the same parameter key is repeated for each value in the array, is known as "multiple values for a single parameter" and it's specified in RFC 3986.

According to the RFC, the query component of a URI can include multiple parameters with the same name by separating each value with an ampersand & character. In other words, the second syntax you mentioned - ?key=1&key=2&key=3 - is a valid way to pass an array of values for a single parameter in an HTTP GET request.

gajddo00:
Do you have a link to that source? 🤔 Because every article I've read says that RFC 3986 contains no mention of this convention and I also can't find it there.

cejanen (approved):
cool
🐫

PR #41, opened 2023-06-06
DocC documentation

gajddo00:
DocC style documentation on the usage of Networking library. Let's do a first round of feedback so I know that the direction I took is okay or figure out what's missing. 🐤

TODO/ADD:
- Document EndpointRequestStorageProcessor in depth


Sources/Networking/Documentation.docc/Documentation.md
```diff
@@ -0,0 +1,297 @@
...
+    responseProcessors: [ResponseProcessing] = [StatusCodeProcessor.shared],
+    errorProcessors: [ErrorProcessing] = []
+)
+```
+
+Adapters and processors are passed during initialisation and cannot be changed afterwards.
+
+The DownloadAPIManager contains a public property that enables you to keep track of current tasks in progress.
+```swift
+var allTasks: [URLSessionDownloadTask] { get async }
```
cejanen, line 150: when seeing this, thinking about better naming, is it equivalent to runningTasks? or does this array contain different  semantic?

Sources/Networking/Documentation.docc/Documentation.md
```diff
@@ -0,0 +1,297 @@
+# Networking
+
+A networking layer using native `UrlSession` and Swift concurrency.
+
```
cejanen, line 4: adding some notes I miss overall:
typealiases we have for responses (tuples)
how storing and reusing stored api calls works (that there is an provider to read it from assets when adding whole folder in original sequence of calls)

Sources/Networking/Documentation.docc/Documentation.md
```diff
@@ -0,0 +1,297 @@
...
+    // custom logic here that determines whether the request should be retried
+    // e.g you can only retry with 5xx http error codes
+}
+```
+
+## Interceptors
+Interceptors are useful pieces of code that modify request/response in the network request pipeline.
+![Interceptors diagram](interceptors-diagram.png)
+
+There are three types you can leverage:<br>
```
cejanen, line 190: For some reason I see "<br>" mark in preview 🤔

Sources/Networking/Documentation.docc/Documentation.md
```diff
@@ -0,0 +1,297 @@
+# Networking
+
+A networking layer using native `UrlSession` and Swift concurrency.
+
+## Overview
+Heavily inspired by Moya, the networking layer's philosophy is focused on creating individual endpoint routers, transforming them into a valid URLRequest objects and applying optional adapters and processors in the network call pipeline.
+
+## Router
+By conforming to the ``Requestable`` protocol, you can define endpoint definitions containing the elementary HTTP request components necessary to create valid HTTP requests.
+<br>**Recommendation:** Follow the `Router` naming convention to explicitly indicate the usage of a router pattern.
```
cejanen, line 10: For some reason I see <br> mark in preview 🤔

Sources/Networking/Documentation.docc/Documentation.md
```diff
@@ -0,0 +1,297 @@
...
+    responseProcessors: [LoggingInterceptor.shared],
+    errorProcessors: [LoggingInterceptor.shared]
+    //
+)
+```
+
+### Authorization
+Networking provides a default authorization handling for OAuth scenarios. Use the default ``AuthorizationTokenInterceptor`` with the APIManager to obtain the behaviour of JWT Bearer authorization header injection and access token expiration refresh flow.
+
+Start by implementing an authorization manager by conforming to ``AuthorizationManaging``. This manager requires you to provide storage defined by ``AuthorizationStorageManaging`` (where OAuth credentials will be stored) and a refresh method that will perform the refresh token network call to obtain a new OAuth pair. Optionally, you can provide custom implementations for ``AuthorizationManaging/authorizeRequest(_:)-6azlk`` (by default, this method sets the authorization header) or access token getter (by default, this method returns the access token saved in provided storage).
```
cejanen, line 224: what is -6azlk?

cejanen, line 224: I would add a note about the fact that if access token is expired or api returns 401 networking library automatically tries to refresh access token

cejanen (changes requested):
Added some notes about this first round, not hard to fix. In common I would more focus on "how it works", what does it do for you automatically, add info about all crucial parts not just classes & protocols

Sources/Networking/Documentation.docc/Documentation.md
```diff
@@ -0,0 +1,297 @@
...
+```swift
+init(
+    urlSession: URLSession = .init(configuration: .default),
+    requestAdapters: [RequestAdapting] = [],
+    responseProcessors: [ResponseProcessing] = [StatusCodeProcessor.shared],
+    errorProcessors: [ErrorProcessing] = []
+)
+```
+
+2. Using custom response provider by conforming to ``ResponseProviding``.
```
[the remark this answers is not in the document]
gajddo00: There is no multipeerconnection modifier... The only mention about it is in the EndpointRequestStorageProcessor and then the Multipeer manager itself but I actually have no idea what it's for so I'd rather someone who authored it document it as well as the ResponseProviding. I can only understand as much as is commented in the actual code if I didn't work on it.

Sources/Networking/Documentation.docc/Documentation.md
```diff
@@ -0,0 +1,297 @@
...
+try await apiManager.request(UserRouter.getUser)
+```
+
+If you specify object type, the APIManager will automatically perform the decoding (given the received JSON correctly maps to the decodable). You can also specify a custom json decoder.
+
+```swift
+let userResponse: UserResponse = try await apiManager.request(UserRouter.getUser)
+```
+
+Provide a custom after failure ``RetryConfiguration``, specifying the count of retries, delay and a handler that determines whether the request should be tried again. Otherwise, ``RetryConfiguration/default`` configuration is used.
```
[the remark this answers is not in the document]
gajddo00: I don't see why not

Sources/Networking/Documentation.docc/Documentation.md
gajddo00: that's just naming for the property link generated by Xcode

Sources/Networking/Documentation.docc/Documentation.md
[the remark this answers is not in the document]
cejanen: 👍

Sources/Networking/Documentation.docc/Documentation.md
gajddo00: agreed, I wouldn't use running as the tasks might be paused, but I'd change it to activeTasks

Sources/Networking/Documentation.docc/Documentation.md
cejanen: my stupidity

cejanen (approved):
✍️

PR #43, opened 2023-06-13
Data and file upload support

tonyskansf:
This pull request introduces an upload feature allowing clients to upload data and files (no stream or multipart file support yet). The implementation follows the existing download feature design with slight deviation whilst aiming to keep consistency and familiarity with the API. With that, the changes also include an example that showcases the usage of the API.

Please, do not consider this implementation as final but rather as the first draft.

## Video

https://github.com/strvcom/ios-networking/assets/20850404/01089bdd-7fc7-4157-ac6e-ecf24c7a88a8

### `UploadTask`

The central component is the `UploadTask` object that encapsulates the ongoing `URLSessionUploadTask` and provides the task's state updates. Each upload request creates an upload task instance that is internally stored until the request is completed. In case of a failed request, the UploadTask serves as the representation of the failed request and is important in subsequent retry attempts.

### `UploadTask.State`

This object represents the progress of the upload request and is updated continuously by the internal publisher. It is then available to the client as an asynchronous sequence. Since the implementation creates the `URLSessionUploadTask` with completion handlers, the response is accessible only upon the task's completion. Consequently, the response or error is delivered within the upload task's state.

### `Retryable`

The `UploadTask` conforms to the `Retryable` protocol instead of the `UploadAPIManager` as is in the `DownloadAPIManager`. This emphasizes the intention that the task itself is retryable, rather than the manager. Despite potentially deviating from the original design of the protocol, this approach seems to work as well. Each upload task has its `retryCounter: Counter` that determines whether the task should be retried based on the `RetryConfiguration`.

## Additional Notes
- The sample does not include a complete router example intentionally. I've used [this service](https://www.filestack.com/) for testing purposes.

### DIFF
--- a/NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
@@ -1,0 +1,42 @@
+//
+//  SampleUploadRouter.swift
+//  NetworkingSampleApp
+//
+//  Created by Tony Ngo on 12.06.2023.
+//
+
+import Foundation
+import Networking
+import UniformTypeIdentifiers
+
+enum SampleUploadRouter: Requestable {
+    case image
+    case file(URL)
+
+    var baseURL: URL {
+        fatalError("Provide your API base URL for upload")
+    }
+
+    var headers: [String : String]? {
+        switch self {
+        case .image:
+            return ["Content-Type": "image/png"]
+        case let .file(url):
+            return ["Content-Type": url.mimeType]
+        }
+    }
+
+    var path: String {
+        fatalError("Provide your API endpoint path for upload")
+    }
+
+    var method: HTTPMethod {
+        .post
+    }
+}
+
+private extension URL {
+    var mimeType: String {
+        UTType(filenameExtension: pathExtension)?.preferredMIMEType ?? "application/octet-stream"
+    }
+}
--- a/NetworkingSampleApp/NetworkingSampleApp/ContentView.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/ContentView.swift
@@ -5,11 +5,13 @@
 //  Created by Matej Molnár on 28.01.2023.
 //
 
+import Networking
 import SwiftUI
 
 enum NetworkingFeature: String, Hashable, CaseIterable {
     case authorization
     case downloads
+    case uploads
 }
 
 struct ContentView: View {
@@ -27,6 +29,12 @@ struct ContentView: View {
                     AuthorizationView()
                 case .downloads:
                     DownloadsView()
+                case .uploads:
+                    UploadsView(viewModel: UploadsViewModel(
+                        uploadService: UploadService(
+                            uploadManager: UploadAPIManager()
+                        )
+                    ))
                 }
             }
         }
--- a/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadItemViewModel.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadItemViewModel.swift
@@ -1,0 +1,42 @@
+//
+//  UploadItemViewModel.swift
+//  NetworkingSampleApp
+//
+//  Created by Tony Ngo on 12.06.2023.
+//
+
+import Foundation
+
+@MainActor
+final class UploadItemViewModel: ObservableObject {
+    @Published private(set) var progress: Double = 0
+    @Published private(set) var formattedProgress: String = ""
+    @Published private(set) var isPaused = false
+    @Published private(set) var isCancelled = false
+    @Published private(set) var isRetryable = false
+
+    let fileName: String
+    let totalProgress = 100.0
+
+    private let item: UploadItem
+    private let uploadService: UploadService
+
+    init(item: UploadItem, uploadService: UploadService) {
+        self.item = item
+        self.fileName = item.fileName
+        self.uploadService = uploadService
+    }
+}
+
+extension UploadItemViewModel {
+    func observeProgress() async {
+        let uploadStateStream = await uploadService.uploadStateStream(for: item.id)
+        for await state in uploadStateStream {
+            progress = state.fractionCompleted * 100
+            formattedProgress = String(format: "%.2f", progress) + "%"
+            isPaused = state.isSuspended
+            isCancelled = state.cancelled
+            isRetryable = state.cancelled || state.timedOut
+        }
+    }
+}
--- a/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadItemViewModel.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadItemViewModel.swift
@@ -15,6 +15,14 @@ final class UploadItemViewModel: ObservableObject {
     @Published private(set) var isCancelled = false
     @Published private(set) var isRetryable = false
 
+    var stateTitle: String {
+        isCancelled
+            ? "Cancelled"
+            : isCompleted ? "Completed" : formattedProgress
+    }
+
+    var isCompleted: Bool { progress == 100 }
+
     let fileName: String
     let totalProgress = 100.0
 
--- a/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadItemViewModel.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadItemViewModel.swift
@@ -36,7 +36,7 @@ extension UploadItemViewModel {
             formattedProgress = String(format: "%.2f", progress) + "%"
             isPaused = state.isSuspended
             isCancelled = state.cancelled
-            isRetryable = state.cancelled || state.timedOut
+            isRetryable = state.cancelled || state.timedOut || state.error != nil
         }
     }
 
@@ -61,6 +61,13 @@ extension UploadItemViewModel {
             await uploadService.cancel(taskId: item.id)
             isCancelled = true
             isRetryable = true
+        }
+    }
+
+    func retry() {
+        Task {
+            try await uploadService.retry(item)
+            await observeProgress()
         }
     }
 }
--- a/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadItemViewModel.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadItemViewModel.swift
@@ -39,4 +39,28 @@ extension UploadItemViewModel {
             isRetryable = state.cancelled || state.timedOut
         }
     }
+
+    func pause() {
+        Task {
+            await uploadService.pause(taskId: item.id)
+            isPaused = true
+            isRetryable = false
+        }
+    }
+
+    func resume() {
+        Task {
+            await uploadService.resume(taskId: item.id)
+            isPaused = false
+            isRetryable = false
+        }
+    }
+
+    func cancel() {
+        Task {
+            await uploadService.cancel(taskId: item.id)
+            isCancelled = true
+            isRetryable = true
+        }
+    }
 }
--- a/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadService.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadService.swift
@@ -24,7 +24,8 @@ extension UploadService {
     func uploadImage(_ data: Data, fileName: String) async throws -> UploadItem {
         let task = try await uploadManager.upload(
             data: data,
-            to: SampleUploadRouter.image
+            to: SampleUploadRouter.image,
+            retryConfiguration: .default
         )
         return UploadItem(
             id: task.id,
@@ -35,7 +36,8 @@ extension UploadService {
     func uploadFile(_ fileUrl: URL) async throws -> UploadItem {
         let task = try await uploadManager.upload(
             fromFile: fileUrl,
-            to: SampleUploadRouter.file(fileUrl)
+            to: SampleUploadRouter.file(fileUrl),
+            retryConfiguration: .default
         )
         return UploadItem(
             id: task.id,
@@ -57,6 +59,13 @@ extension UploadService {
 
     func cancel(taskId: String) async {
         await uploadManager.task(with: taskId)?.cancel()
+    }
+
+    func retry(_ uploadItem: UploadItem) async throws {
+        try await uploadManager.retry(
+            taskId: uploadItem.id,
+            retryConfiguration: .default
+        )
     }
 }
 
--- a/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadService.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadService.swift
@@ -46,6 +46,18 @@ extension UploadService {
     func uploadStateStream(for uploadTaskId: String) async -> UploadAPIManaging.StateStream {
         await uploadManager.stateStream(for: uploadTaskId)
     }
+
+    func pause(taskId: String) async {
+        await uploadManager.task(with: taskId)?.pause()
+    }
+
+    func resume(taskId: String) async {
+        await uploadManager.task(with: taskId)?.resume()
+    }
+
+    func cancel(taskId: String) async {
+        await uploadManager.task(with: taskId)?.cancel()
+    }
 }
 
 private extension UploadAPIManaging {
--- a/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadService.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadService.swift
@@ -1,0 +1,55 @@
+//
+//  UploadService.swift
+//  NetworkingSampleApp
+//
+//  Created by Tony Ngo on 12.06.2023.
+//
+
+import Foundation
+import Networking
+
+final class UploadService {
+    private let uploadManager: UploadAPIManaging
+
+    init(uploadManager: UploadAPIManaging) {
+        self.uploadManager = uploadManager
+    }
+
+    deinit {
+        uploadManager.invalidateSession(shouldFinishTasks: false)
+    }
+}
+
+extension UploadService {
+    func uploadImage(_ data: Data, fileName: String) async throws -> UploadItem {
+        let task = try await uploadManager.upload(
+            data: data,
+            to: SampleUploadRouter.image
+        )
+        return UploadItem(
+            id: task.id,
+            fileName: fileName
+        )
+    }
+
+    func uploadFile(_ fileUrl: URL) async throws -> UploadItem {
+        let task = try await uploadManager.upload(
+            fromFile: fileUrl,
+            to: SampleUploadRouter.file(fileUrl)
+        )
+        return UploadItem(
+            id: task.id,
+            fileName: fileUrl.lastPathComponent
+        )
+    }
+
+    func uploadStateStream(for uploadTaskId: String) async -> UploadAPIManaging.StateStream {
+        await uploadManager.stateStream(for: uploadTaskId)
+    }
+}
+
+private extension UploadAPIManaging {
+    func task(with id: String) async -> UploadTask? {
+        await allTasks.first { $0.id == id }
+    }
+}
--- a/Sources/Networking/Core/Upload/UploadAPIManager.swift
+++ b/Sources/Networking/Core/Upload/UploadAPIManager.swift
@@ -224,15 +224,7 @@ private extension UploadAPIManager {
             task.resume()
             return uploadTask
         } catch {
-            do {
-                return try await uploadRequest(
-                    uploadable,
-                    request: request,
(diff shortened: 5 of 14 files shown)


NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
```diff
@@ -0,0 +1,42 @@
...
+
+enum SampleUploadRouter: Requestable {
+    case image
+    case file(URL)
+
+    var baseURL: URL {
+        fatalError("Provide your API base URL for upload")
+    }
+
+    var headers: [String : String]? {
```
brenovaladao, line 20:
```suggestion
    var headers: [String: String]? {
```

NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
```diff
@@ -0,0 +1,42 @@
...
+        switch self {
+        case .image:
+            return ["Content-Type": "image/png"]
+        case let .file(url):
+            return ["Content-Type": url.mimeType]
+        }
+    }
+
+    var path: String {
+        fatalError("Provide your API endpoint path for upload")
```
brenovaladao, line 30: do we want a functional sample? or just code wise?

tonyskansf: If it's desired I can add httpbin.org which I was using in dev.

brenovaladao: got it, I wouldn't say it's needed for just demonstration - but final call is not on me

Sources/Networking/Core/Upload/UploadAPIManager.swift
```diff
@@ -0,0 +1,286 @@
...
+//  Created by Tony Ngo on 12.06.2023.
+//
+
+import Combine
+import Foundation
+
+/// Default upload API manager
+open class UploadAPIManager: NSObject {
+    // MARK: - Public Properties
+    public var allTasks: [UploadTask] {
```
gajddo00, line 14: We decided to go with activeTasks instead of allTasks for the DownloadAPIManager as well (in the docc PR for now).

Sources/Networking/Core/Upload/UploadAPIManager.swift
```diff
@@ -0,0 +1,286 @@
...
+        with uploadable: Uploadable,
+        for request: URLRequest,
+        completionHandler: @escaping @Sendable (Data?, URLResponse?, Error?) -> Void
+    ) -> URLSessionUploadTask {
+        switch uploadable {
+        case let .data(data):
+            return urlSession.uploadTask(
+                with: request,
+                from: data,
+                completionHandler: completionHandler
```
gajddo00, line 263: why not use the async version?

Sources/Networking/Core/Upload/UploadAPIManager.swift
```diff
@@ -0,0 +1,286 @@
...
+                        retryConfiguration: retryConfiguration
+                    )
+                } catch {
+                    state.error = await errorProcessors.process(
+                        error,
+                        for: uploadTask.endpointRequest
+                    )
+
+                    // No cleanup in case the task will be retried.
+                    try await uploadTask.complete(with: state)
```
gajddo00, line 247: If the user never retries by using the retry method, will the task ever get removed from the uploadTasks if it fails?

Sources/Networking/Core/Upload/UploadAPIManager.swift
tonyskansf: Good question! 🤔 We could definitely consider using the async version, however, if we go down that route, we'll need to make some slight changes to the implementation, because:
- The async versions don't return `URLSessionTask` and will be suspended, which means returning immediately as it is now wouldn't be possible. So we'll need to consider what to return to the client, if anything at all.
- We'll need to handle errors and responses from the request using delegates.

If you have any ideas or suggestions, please feel free to share them. 🙌

Sources/Networking/Core/Upload/UploadAPIManager.swift
tonyskansf: Currently only when the manager is deinitialized. What do you think would be the best to handle this?

Sources/Networking/Core/Upload/UploadAPIManager.swift
gajddo00: Ah, that's unfortunate but I don't see a way to make it work as conveniently with the async version, as you're saying. Let's just keep it as is :)

Sources/Networking/Core/Upload/UploadAPIManager.swift
gajddo00: There are only two ways I guess, either we have retry functionality as is but keep hoarding failed requests (which in case of singletons is not great) or we don't have the retry function and just remove it from uploadTasks after it fails. :(

gajddo00:
Overall, it's nice 🍓 I just don't know how I feel about the fact that conceptually it's 95% same logic as the DownloadAPIManager but the implementation is entirely different, so now we have two managers that do almost the same thing but very differently. 🤔

Sources/Networking/Core/Upload/UploadAPIManager.swift
tonyskansf: We can explore an alternative approach by wrapping the asynchronous part of the request in a `Task`, which can be returned in the `UploadTask` object (this would give us some control over the lifetime of that Task too). However, as you mentioned, this approach may not be as convenient as using the completion closure.

Sources/Networking/Core/Upload/UploadAPIManager.swift
tonyskansf: When implementing, I had some doubts regarding the expected retry behavior for uploads. Is it even common practice to automatically retry failed uploads? For instance, if an upload request fails in the middle of the process and is retried without providing any feedback to the user, it might appear strange that the progress suddenly jumps from 80% back to 10%. 

Considering this case, it might be better to leave the retry behavior as the dev's responsibility. From the library's perspective, any upload request that errors out would be canceled (as you mentioned). This approach allows clients/devs to have more control over the retry logic according to their specific requirements.

NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
cejanen: I would add some default value, I don't feel it's a good idea throw fatal error when showcasing sample solution

Sources/Networking/Core/Upload/UploadAPIManager.swift
cejanen: Agree, I would keep it as it is. Even though async/await is cool, still older approaches (closures, combine) still their own purposes and I don't mind mixing them if it makes sense. I would just consider to put @tonyskansf explanation into code as a comment. Best place may be the protocol where this method is defined.

Sources/Networking/Core/Upload/UploadAPIManager.swift
cejanen: Agree with @tonyskansf , would be very helpful to have retry functionality in place, at the same time it's true we don't know use case behaviour. 
When uploading in chunks retry should continue with last successful chunk +1 etc. 
For sake of simplicity I would go with option that upload doesn't have retry at all so it's dev responsibility eventually

Sources/Networking/Core/Upload/UploadTask.swift
```diff
@@ -0,0 +1,122 @@
...
+    var task: URLSessionUploadTask
+
+    /// The request associated with this task.
+    let endpointRequest: EndpointRequest
+
+    /// The uploadable data associated with this task.
+    let uploadable: Uploadable
+
+    /// Use this publisher to emit a new state of the task.
+    let statePublisher: CurrentValueSubject<State, Never>
```
cejanen, line 25: just throwing idea here... 
this property is publicly used just to pass new value from URLSessionTaskDelegate... in all other cases sending values is extension below. What if we would change this accessory level to private and make public method to pass value into it? similarly we have other methods in extension?

Sources/Networking/Core/Upload/UploadTask+State.swift
```diff
@@ -0,0 +1,69 @@
+//
+//  UploadTask+State.swift
+//  
+//
+//  Created by Tony Ngo on 12.06.2023.
+//
+
+import Foundation
+
+extension UploadTask {
```
cejanen, line 10: Should be this extension also public?

Sources/Networking/Core/Upload/UploadTask+State.swift
```diff
@@ -0,0 +1,69 @@
...
+        public let totalBytes: Int64
+
+        /// An error produced by the task.
+        public var error: Error?
+
+        /// A response produced by the task.
+        public var response: Response?
+
+        /// The internal state of the `URLSessionTask`.
+        let taskState: URLSessionTask.State
```
cejanen, line 26: why this is not public? can it be private in other case?

Sources/Networking/Core/Upload/UploadAPIManaging.swift
```diff
@@ -0,0 +1,60 @@
...
+    /// i.e., `UploadTask.State.error` is non-nil. In such case, you can call `retry(taskId:)` to re-activate the stream for the specified `uploadTaskId`.
+    /// - Parameter uploadTaskId: The identifier of the task to observe.
+    /// - Returns: An asynchronous stream of upload state. If there is no such upload task the return stream finishes immediately.
+    func stateStream(for uploadTaskId: UploadTask.ID) async -> StateStream
+
+    /// Invalidates the session with the option to wait for all outstanding (active) tasks.
+    ///
+    /// The internal implementation uses Apple's delegate pattern which retains a strong reference to the delegate. You must call this method to allow the manager to be released from the memory, otherwise your app will be leaking until your app exits or the session is invalidated.
+    /// - Parameter shouldFinishTasks: Determines whether all outstanding tasks should finish before invalidating the session or be immediately cancelled.
+    func invalidateSession(shouldFinishTasks: Bool)
```
cejanen, line 59: mentioned this before somewhere to add usage explanation... I'm ok with this one

cejanen (changes requested):
Overall it's good move... besides some minor requests from me I would like to ask
@gajddo00 to mention concrete examples of different approaches between download & upload
we should figure out whether to keep similar retry mechanism for download/upload as it is in APIManager or go with option of retry method for failed download/upload task @brenovaladao @johnkodes @arkhigleb ?

NetworkingSampleApp/NetworkingSampleApp/ContentView.swift
```diff
@@ -27,6 +29,12 @@ struct ContentView: View {
                     AuthorizationView()
                 case .downloads:
                     DownloadsView()
+                case .uploads:
+                    UploadsView(viewModel: UploadsViewModel(
```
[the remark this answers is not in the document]
tonyskansf: Moved everything inside the view and view model, while keeping the service so the import of the library is only there.

NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
```diff
@@ -0,0 +1,42 @@
...
+import Foundation
+import Networking
+import UniformTypeIdentifiers
+
+enum SampleUploadRouter: Requestable {
+    case image
+    case file(URL)
+
+    var baseURL: URL {
+        fatalError("Provide your API base URL for upload")
```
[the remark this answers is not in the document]
tonyskansf: I'm not sure what you mean (missing comment about the path?). 😄

NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    var method: HTTPMethod {
+        .post
+    }
+}
+
+private extension URL {
+    var mimeType: String {
+        UTType(filenameExtension: pathExtension)?.preferredMIMEType ?? "application/octet-stream"
```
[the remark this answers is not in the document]
tonyskansf: As a sample, it does not matter as much, IMO. In the actual use case, it would be preferred to have the mime type provided by the caller — something like `case file(URL, mimeType: String)`.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadService.swift
```diff
@@ -0,0 +1,76 @@
...
+    func retry(_ uploadItem: UploadItem) async throws {
+        try await uploadManager.retry(
+            taskId: uploadItem.id,
+            retryConfiguration: .default
+        )
+    }
+}
+
+private extension UploadAPIManaging {
+    func task(with id: String) async -> UploadTask? {
```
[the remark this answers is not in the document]
tonyskansf: It wouldn't do any harm to add it. I think the only concern would be an expectation of correctly implementing the `activeTasks` variable by the client's concrete `UploadAPIManaging` implementation. 🤔 Let me add it anyway.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadItemViewModel.swift
```diff
@@ -0,0 +1,81 @@
...
+            isPaused = state.isSuspended
+            isCancelled = state.cancelled
+            isRetryable = state.cancelled || state.timedOut || state.error != nil
+        }
+    }
+
+    func pause() {
+        Task {
+            await uploadService.pause(taskId: item.id)
+            isPaused = true
```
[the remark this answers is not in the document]
tonyskansf: Do you mean the `isPaused` property? This was from me just to make sure that the view model is in the correct state, otherwise the upload task's state is available in its stream here — a new state is available whenever the upload task's state changes, in this case being paused/suspended.

```swift
for await state in uploadStateStream {
    isPaused = state.isSuspended
}
```

Sources/Networking/Core/Upload/UploadAPIManager.swift
```diff
@@ -0,0 +1,286 @@
...
+                .values
+                // Values may contain inactive tasks
+                .filter { activeTasks.contains($0.task) }
+        }
+    }
+
+    // MARK: - Private Properties
+    private var uploadTasks = ThreadSafeDictionary<String, UploadTask>()
+
+    private lazy var urlSession = URLSession(
```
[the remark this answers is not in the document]
tonyskansf: Sure. I prefer keeping the property lazily initialized, wdyt?

Sources/Networking/Core/Upload/UploadTask+State.swift
tonyskansf: Making this extension wouldn't change anything. Although, I can mark it as public and remove the public access level from the `UploadTask.State` on line 12.

Sources/Networking/Core/Upload/UploadTask+State.swift
tonyskansf: It's an internal implementation detail that I didn't want to expose to the client. It's not private so it can be accessible in potential tests.

Sources/Networking/Core/Upload/UploadAPIManager.swift
tonyskansf: > When uploading in chunks retry should continue with last successful chunk +1 etc.

I believe this depends on the server's implementation too — if it supports partial (resumable) upload.

gajddo00:
@cejanen What I mean was just different approaches to the same thing which is absolutely fine, but most of the functionality is the same so I was thinking whether it would be feasible to have one generic solution for both that could be inherited from or something in that sense. 
Because the differences come down to
- urlSession.download vs urlSession.upload
- UrlSessionDownloadTask vs URLSessionUploadTask
- different progress state object emitted in the stream
🤔 but I'm just thinking out loud here, maybe it wouldn't even be possible

Sources/Networking/Core/Upload/UploadTask.swift
tonyskansf: The publisher is accessible only within the library, not publicly, whilst the methods in the extension below are. 

> What if we would change this accessory level to private and make public method to pass value into it? similarly we have other methods in extension?

I don't think allowing the client to send new states is a good idea. This publisher should stay an internal implementation detail IMO. Wdyt?

tonyskansf:
Some things (objects) could certainly be unified, but I'd rather not decide this as it changes the implementation of either of the managers.

NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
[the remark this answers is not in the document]
cejanen: don't throw fatal error instead having default value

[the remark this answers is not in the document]
tonyskansf: I've added a sample server for trying post requests.

brenovaladao:
got an error when uploading a specific photo from the simulator

> Error Domain=NSItemProviderErrorDomain Code=-1000 "Cannot load representation of type public.jpeg" UserInfo={NSLocalizedDescription=Cannot load representation of type public.jpeg, NSUnderlyingError=0x6000036ba9a0 {Error Domain=NSCocoaErrorDomain Code=4101 "Couldn’t communicate with a helper application." UserInfo={NSUnderlyingError=0x6000036bdda0 {Error Domain=PHAssetExportRequestErrorDomain Code=0 "The operation couldn’t be completed. (PHAssetExportRequestErrorDomain error 0.)" UserInfo={NSLocalizedDescription=The operation couldn’t be completed. (PHAssetExportRequestErrorDomain error 0.), NSUnderlyingError=0x6000036bf660 {Error Domain=PAMediaConversionServiceErrorDomain Code=2 "The operation couldn’t be completed. (PAMediaConversionServiceErrorDomain error 2.)" UserInfo=0x6000038e1820 (not displayed)}}}}}}

gajddo00:
I also got this, but it's the simulator problem with corrupted images, nothing to do with uploading I think

cejanen (commented):
@tonyskansf  I answered last comments or resolved issues, we're almost done 👍

Sources/Networking/Core/Upload/UploadAPIManager.swift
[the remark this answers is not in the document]
cejanen: Yes, my point was of using session config from init eventually

Sources/Networking/Core/Upload/UploadAPIManager.swift
```diff
@@ -0,0 +1,293 @@
...
+        response: URLResponse?,
+        error: Error?
+    ) {
+        Task {
+            guard let uploadTask = await uploadTasks.getValue(for: endpointRequest.id) else {
+                return
+            }
+
+            var state = UploadTask.State(task: uploadTask.task)
+            if let data, let response {
```
gajddo00, line 210: Since the returned task is not used, just called, the throwing statements inside are swallowed and the error is never propagated. It happened to us since we were receiving a 500 for a bad form and we never received any errors (it never hit the error processor).

Sources/Networking/Core/Upload/UploadAPIManager.swift
```diff
@@ -0,0 +1,293 @@
...
+    ///   as it is now wouldn't be possible. So we'll need to consider what to return to the client, if anything at all.
+    /// - We'll need to handle errors and responses from the request using delegates.
+    func sessionUploadTask(
+        with uploadable: Uploadable,
+        for request: URLRequest,
+        completionHandler: @escaping @Sendable (Data?, URLResponse?, Error?) -> Void
+    ) -> URLSessionUploadTask {
+        switch uploadable {
+        case let .data(data):
+            return urlSession.uploadTask(
```
gajddo00, line 267: Thread 12: "Completion handler blocks are not supported in background sessions. Use a delegate instead."

Due to the use of completion blocks, UploadAPIManager cannot be used in background!!

gajddo00: https://github.com/strvcom/ios-networking/pull/48

Sources/Networking/Core/Upload/UploadAPIManager.swift
gajddo00: https://github.com/strvcom/ios-networking/pull/48

Sources/Networking/Core/Upload/UploadAPIManager.swift
```diff
@@ -0,0 +1,293 @@
...
+        } else {
+            urlSession.invalidateAndCancel()
+        }
+    }
+
+    public func upload(
+        data: Data,
+        to endpoint: Requestable,
+        retryConfiguration: RetryConfiguration?
+    ) async throws -> UploadTask {
```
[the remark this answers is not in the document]
cejanen: Not for now, I'll open ticket in Jira for that. I don't see simple solution now as we're returning data wrapped in UploadTask.

cejanen (approved):
🥡

PR #44, opened 2023-06-19
Add `multipart/form-data` support

tonyskansf:
This pull request adds `multipart/form-data` content type upload support on top of the existing functionality in #43. Also, the changes contain an example showcasing the usage of the API.

### `MultipartFormDataEncoder`

Provides two ways of encoding the MultipartFormData: either in-memory encoding or encoding to disk. In-memory encoding is suitable for small payloads while encoding to disk is preferred and more efficient for large payloads to prevent memory issues. 

Regardless of the encoding type, all data are encoded into this representation:

```
--boundary-{some id}
Content-Disposition: form-data; name="field-name"

{Field data}
--boundary-{some id}
Content-Disposition: form-data; name="attachment"; filename="file.txt"
Content-Type: text/plain

{Contents of the file}
--boundary-{some id}--
```

## Usage

```swift
let uploadManager = UploadAPIManager()

// Get form data from the user (multipart form body parts)
let textFieldData = Data(text.utf8)
let attachmentUrl = URL(string: "path/to/file")!

// Add them to the MultipartFormData instance
let multipartFormData = MultipartFormData(/* boundary: override boundary */)
multipartFormData.append(textFieldData, name: "text-field")
try multipartFormData.append(from: attachmentUrl, name: "attachment-field")

// Send request
try await uploadManager.upload(
  multipartFormData: multipartFormData,
  to: endpoint,
  retryConfiguration: .default
)
```


NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/FormUploadsViewModel.swift
```diff
@@ -0,0 +1,55 @@
...
+
+                uploadItemViewModels.append(UploadItemViewModel(
+                    item: uploadItem,
+                    uploadService: uploadService
+                ))
+
+                text = ""
+                fileUrl = nil
+            } catch {
+                print("Failed to upload with error:", error)
```
brenovaladao, line 51: I think I saw some error binding in other examples in this repo, if true, maybe we can use it here as well

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadsView.swift
```diff
@@ -61,3 +47,88 @@ struct UploadsView: View {
...
+                    matching: .images
+                )
+                .onChange(of: selectedPhotoPickerItem) { photo in
+                    Task {
+                        if let data = try? await photo?.loadTransferable(type: Data.self) {
+                            await viewModel.uploadImage(
+                                data,
+                                fileName: selectedPhotoPickerItem?.supportedContentTypes.first?.preferredFilenameExtension
+                            )
+                        }
```
brenovaladao, line 67: I'd put this piece of code in the VM, sending the photo as param

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadsView.swift
```diff
@@ -61,3 +47,88 @@ struct UploadsView: View {
...
+            Button("File") { isFileImporterPresented = true }
+                .fileImporter(
+                    isPresented: $isFileImporterPresented,
+                    allowedContentTypes: [.mp3, .mpeg4Movie]
+                ) { result in
+                    Task {
+                        if let fileUrl = try? result.get() {
+                            await viewModel.uploadFile(at: fileUrl)
+                        }
+                    }
```
brenovaladao, line 80: same here

Sources/Networking/Core/Upload/MultipartFormData+BodyPart.swift
```diff
@@ -0,0 +1,49 @@
...
+
+        /// An optional file parameter of the `Content-Disposition` header field. This value may be provided if the body part represents a content of a file.
+        let fileName: String?
+
+        /// An optional value of the `Content-Type` header field.
+        let mimeType: String?
+    }
+}
+
+extension MultipartFormData.BodyPart {
```
brenovaladao, line 30: doesn't this needs to be public as well?

Sources/Networking/Core/Upload/MultipartFormDataEncoder.swift
brenovaladao, line 1: why do we need to save the file? sort of cache for a retry?

brenovaladao (commented):
🪙

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadsView.swift
tonyskansf: This has been changed in the base branch based on your comment there. It's not been merged here yet.

tonyskansf: This has been changed in the base branch based on your comment there. It's not been merged here yet.

Sources/Networking/Core/Upload/MultipartFormData+BodyPart.swift
tonyskansf: The `BodyPart` can be only created when appending data to the `MultipartFormData` and because the current implementation doesn't provide any API to work with body parts directly it wasn't needed to expose this detail to the client. 

I'm sure the implementation doesn't cover all use cases so if necessary we can certainly make it public.

Sources/Networking/Core/Upload/MultipartFormDataEncoder.swift
tonyskansf: The answer is similar to why there are [`uploadTask(with:from:)`](https://developer.apple.com/documentation/foundation/urlsession/1409763-uploadtask) and [`uploadTask(with:fromFile:)`](https://developer.apple.com/documentation/foundation/urlsession/1411550-uploadtask) ig one could convert the content of the file to `Data` using `Data(contentsOf:)` and use the first method to send data.

One has the data available in memory while the second reads the data directly from the file thus doesn't load the data into memory so it is more efficient.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadsView.swift
```diff
@@ -48,7 +30,112 @@ struct UploadsView: View {
...
+                    VStack {
+                        ForEach(formViewModel.uploadItemViewModels.indices, id: \.self) { index in
+                            let viewModel = formViewModel.uploadItemViewModels[index]
+                            UploadItemView(viewModel: viewModel)
+                        }
+                    }
+                }
+            }
         }
+        .alert(
```
cejanen, line 47: for me it's a bit weirdo having two view models here, but I believe for sample app it's ok

cejanen (changes requested):
I left some comments in first round

Sources/Networking/Core/Upload/MultipartFormData+BodyPart.swift
tonyskansf: I always thought it was the other way around. I'll change it.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/FormUploadsViewModel.swift
```diff
@@ -0,0 +1,59 @@
...
```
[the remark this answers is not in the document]
tonyskansf: Updated

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/FormUploadsViewModel.swift
```diff
@@ -0,0 +1,59 @@
...
+                        try form.append(from: fileUrl, name: "attachment")
+                    }
+                }
+
+                uploadItemViewModels.append(UploadItemViewModel(
+                    item: uploadItem,
+                    uploadService: uploadService
+                ))
+
+                text = ""
```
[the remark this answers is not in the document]
tonyskansf: It's just meant to show that we can add different types to the multipart upload, apart from files. It's the text bound to the text field in the sample UI. Can be literally anything, any name.

I've renamed it to "username".

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadService.swift
```diff
@@ -45,6 +45,20 @@ extension UploadService {
         )
     }
 
+    func uploadFormData(_ build: @escaping (MultipartFormData) throws -> Void) async throws -> UploadItem {
```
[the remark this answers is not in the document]
tonyskansf: It's just an example of how a `MultipartFormData` object can be constructed and how body parts can be appended to it.

I don't see why the building part of the form data should be asynchronous.

Do you mind sharing your thoughts on how this API might look like?

NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
```diff
@@ -23,6 +24,8 @@ enum SampleUploadRouter: Requestable {
             return ["Content-Type": "image/png"]
         case let .file(url):
             return ["Content-Type": url.mimeType]
+        case let .multipart(boundary):
+            return ["Content-Type": "multipart/form-data; boundary=\(boundary)"]
         }
     }
```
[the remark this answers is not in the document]
tonyskansf: You're totally right! I'll update it.

Sources/Networking/Core/Upload/MultipartFormDataEncoder.swift
```diff
@@ -0,0 +1,183 @@
+//
+//  MultipartFormDataEncoder.swift
+//  
+//
+//  Created by Tony Ngo on 18.06.2023.
+//
+
+import Foundation
+
+open class MultipartFormDataEncoder {
```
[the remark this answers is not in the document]
tonyskansf: It's greatly inspired by Alamofire.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadsView.swift
tonyskansf: Agreed. I could've given more love to the UI part of the example. 😞

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/FormUploadsViewModel.swift
```diff
@@ -0,0 +1,59 @@
...
+    @Published var fileUrl: URL?
+    @Published var isErrorAlertPresented = false
+    @Published private(set) var error: Error?
+    @Published private(set) var uploadItemViewModels: [UploadItemViewModel] = []
+
+    var selectedFileName: String {
+        let resources = try? fileUrl?.resourceValues(forKeys:[.fileSizeKey])
+        let fileSize = (resources?.fileSize ?? 0) / 1_000_000
+        var fileName = fileUrl?.lastPathComponent ?? ""
+        if fileSize > 0 { fileName += "\n\(fileSize) MB" }
```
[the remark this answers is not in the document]
tonyskansf: Didn't think about it. Thanks — there's a `ByteCountFormatter`.

NetworkingSampleApp/NetworkingSampleApp/Extensions/ByteCountFormatter+Convenience.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+
+import Foundation
+
+extension ByteCountFormatter {
+    static let megaBytesFormatter: ByteCountFormatter = {
+        let formatter = ByteCountFormatter()
+        formatter.allowedUnits = [.useMB]
+        formatter.countStyle = .file
+        return formatter
```
brenovaladao, line 15: nice

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadService.swift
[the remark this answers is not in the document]
cejanen: Sorry mentioning async/await is confusing. My idea is if we can have something like
```
    func uploadFormData(data: MultipartFormData) async throws -> UploadItem
    ```
the method is called here
```
let uploadItem = try await uploadService.uploadFormData { form in
                    form.append(Data(self.username.utf8), name: "username-textfield")

                    if let fileUrl = self.fileUrl {
                        try form.append(from: fileUrl, name: "attachment")
                    }
                }
 ```
 what if we create build data method in FormUploadsViewModel and have
 let data = buildMultipartData()
 uploadFormData(data)

originally in uploadData method there are two lines which may be simply moved
let multipartFormData = MultipartFormData()
try build(multipartFormData)

I feel we don't need closure here and if there is any async work we can use async/await

Sources/Networking/Core/Upload/UploadTask.swift
```diff
@@ -26,8 +26,12 @@ public struct UploadTask {
 
     /// The counter that counts number of retries for this task.
     let retryCounter: Counter
+
+    /// The file manager associated with the task.
+    let fileManager: FileManager
```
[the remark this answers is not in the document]
cejanen: Honestly, no clue. But I believe that FileManager should always keep integrity when working with filesystem. So if you have various file manager instances and want to remove file any of them may remove that file (even different way internally)

cejanen (commented):
One thing left on my side...  @tonyskansf 
otherwise we're done here 👍

NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUploadRouter.swift
```diff
@@ -23,6 +23,8 @@ enum SampleUploadRouter: Requestable {
             return ["Content-Type": "image/png"]
         case let .file(url):
             return ["Content-Type": url.mimeType]
+        case let .multipart(boundary):
+            return ["Content-Type": "multipart/form-data; boundary=\(boundary)"]
```
gajddo00, line 27: We needed to drop the first two -- otherwise the (NodeJS) server kept returning "Unexpected end of form". Are there different boundary standards across servers or is this a bug on our side?

Sources/Networking/Core/Upload/MultipartFormDataEncoder.swift
cejanen: Can we just leave this information in any kind of documentation?

PR #47, opened 2023-07-13
[FIX] Download file location no longer exists

gajddo00:
The location returned from DownloadAPIManager no longer existed because of async progress updates (the location only temporarily exists in the scope of the method).
fix --> move downloaded file into documents to persist the downloaded file for further use.

### DIFF
--- a/Sources/Networking/Core/DownloadAPIManager.swift
+++ b/Sources/Networking/Core/DownloadAPIManager.swift
@@ -227,17 +227,17 @@ extension DownloadAPIManager: URLSessionDelegate, URLSessionDownloadDelegate {
 
 extension URL {
     enum FileError: Error {
-        case documentsDirUnavailable
+        case documentsDirectoryUnavailable
     }
     
     func moveContentsToDocuments(response: URLResponse) throws -> URL {
         guard let documentsURL = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first else {
-            throw FileError.documentsDirUnavailable
+            throw FileError.documentsDirectoryUnavailable
         }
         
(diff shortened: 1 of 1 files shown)


Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -212,3 +224,23 @@ extension DownloadAPIManager: URLSessionDelegate, URLSessionDownloadDelegate {
         }
     }
 }
+
+extension URL {
+    enum FileError: Error {
+        case documentsDirUnavailable
```
matejmolnar, line 230: `documentDirectoryUnavailable` would probably be a better name

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -212,3 +224,23 @@ extension DownloadAPIManager: URLSessionDelegate, URLSessionDownloadDelegate {
...
+    }
+    
+    func moveContentsToDocuments(response: URLResponse) throws -> URL {
+        guard let documentsURL = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first else {
+            throw FileError.documentsDirUnavailable
+        }
+        
+        // Use original extension, otherwise urlSession saves file as .tmp
+        let pathExtension = response.url?.pathExtension ?? pathExtension
+        let newURL = documentsURL.appendingPathComponent("temp.\(pathExtension)")
```
matejmolnar, line 240: Can we rather get the actual file name instead of naming it `temp`?

matejmolnar (approved):

matejmolnar (approved):
Looks better, thanks 🙂

PR #53, opened 2023-12-06
[feat] Downloads UI updates

matejmolnar:
## Changes
- Update the Downloads screen UI to be similar to the Uploads UI
- On appear load active download tasks

https://github.com/strvcom/ios-networking/assets/17743607/075ddc20-1cef-4eb5-b169-55aa348b8ef4

### DIFF
--- a/NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressView.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressView.swift
@@ -44,7 +44,7 @@ private extension DownloadProgressView {
                         .progressViewStyle(.linear)
                         .frame(width: 150)
 
-                    Text("\(String(format: "%.1f", viewModel.state.megaBytesCompleted))MB")
+                    Text(viewModel.state.downloadedBytes)
                         .font(.footnote)
                         .foregroundColor(.gray)
 
--- a/NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressView.swift
+++ b/NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressView.swift
@@ -50,37 +50,17 @@ private extension DownloadProgressView {
 
                     Spacer()
 
-                    button(
-                        symbol: viewModel.state.status == .suspended ? "play" : "pause",
-                        color: .blue,
-                        action: { viewModel.state.status == .suspended ? viewModel.resume() : viewModel.suspend() }
-                    )
+                    TaskButton(config: viewModel.state.status == .suspended ? .play : .pause) {
+                        viewModel.state.status == .suspended ? viewModel.resume() : viewModel.suspend()
+                    }
 
-                    button(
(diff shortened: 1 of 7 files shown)


cejanen (changes requested):
Nice finding and proactive step. Two minor things worth to look at.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressView.swift
```diff
@@ -11,65 +11,76 @@ struct DownloadProgressView: View {
...
-                    viewModel.cancel()
-                } label: {
-                    Text("Cancel")
+
+                HStack {
+                    ProgressView(value: viewModel.state.percentCompleted, total: 100)
+                        .progressViewStyle(.linear)
+                        .frame(width: 150)
+
+                    Text("\(String(format: "%.1f", viewModel.state.megaBytesCompleted))MB")
```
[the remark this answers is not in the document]
matejmolnar: Good idea, I didn't notice that 👍

NetworkingSampleApp/NetworkingSampleApp/UIElements/TaskButton.swift
```diff
@@ -0,0 +1,54 @@
...
+import SwiftUI
+
+struct TaskButton: View {
+    enum Config {
+        case play, pause, cancel, retry
+
+        var imageName: String {
+            switch self {
+            case .play:
+                return "play"
```
cejanen, line 17: do we need return here?

cejanen: similar for all switches about Self?

cejanen (changes requested):
One more comment, if it's possible to update

NetworkingSampleApp/NetworkingSampleApp/UIElements/TaskButton.swift
matejmolnar: Done ✅

NetworkingSampleApp/NetworkingSampleApp/UIElements/TaskButton.swift
```diff
@@ -0,0 +1,50 @@
...
+            switch self {
+            case .play, .pause, .retry:
+                .blue
+            case .cancel: 
+                .red
+            }
+        }
+    }
+
+    let config: Config
```
cejanen, line 33: I'm always challenging the public properties even though let. Does it make sense for any other structures to have those two visible?

cejanen (dismissed):
I left comment which is more subjective so we're good to go.

🍝

NetworkingSampleApp/NetworkingSampleApp/UIElements/TaskButton.swift
matejmolnar: True, I made them private.

cejanen (approved):
🦧

johnkodes (approved):

Issue #39, opened 2023-06-05
More samples


cejanen:
Would be nice to have in total:

simple get request
multiple paralel requests
post request
download data(image, file)
upload data
authentication(login) + refresh token

PR #54, opened 2023-12-11
[feat] Users screen with more samples

matejmolnar:
This pull request is aiming to solve issue #39 

## Changes
- Add Users scene which utilises `SampleUserRouter`.
- Add showcase of a GET request by fetching/displaying multiple users serially or parallelly.
- Add showcase of a POST request by creating a new user with a custom `responseDecoder` due to the `dateDecodingStrategy`.

https://github.com/strvcom/ios-networking/assets/17743607/8c70c944-e712-4337-9aa0-86b8292f3937


NetworkingSampleApp/NetworkingSampleApp/Scenes/Users/UsersViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+            responseProcessors: responseProcessors,
+            errorProcessors: errorProcessors
+        )
+    }()
+}
+
+extension UsersViewModel {
+    func getUsers(in range: ClosedRange<Int>, parallelFetch: Bool) {
+        Task {
+            users = []
```
cejanen, line 48: do we need to reset users like this? or is it different from users.removeAll()? Do we need to call it in task or can it be before the task starts?

cejanen (commented):
I left one question which I would like to know. Other we're good here

NetworkingSampleApp/NetworkingSampleApp/Scenes/Users/UsersViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+//  NetworkingSampleApp
+//
+//  Created by Matej Molnár on 07.12.2023.
+//
+
+import Foundation
+import Networking
+
+@MainActor
+final class UsersViewModel: ObservableObject {
```
johnkodes, line 12: @matejmolnar @cejanen Why do we sometimes use a service as well in the VMs and sometimes we don't? 

Eg. the upload also has a `UploadService` 😅

cejanen: Usually for samples we don't. For samples we try to keep code as simple as possible just to showcase the library usage. Because upload service logic is shared in multiple VMs it made sense to extract it own service class.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Users/UsersViewModel.swift
matejmolnar: No it doesn't matter whether we do it inside the Task or with the `removeAll()` method. What is your preference?

NetworkingSampleApp/NetworkingSampleApp/Scenes/Users/UsersViewModel.swift
matejmolnar: Although I am still not entirely sure why is the UploadAPIManager implementation so different and more complex than the DownloadAPIManager 🤔 but that's a topic for another debate 😄

cejanen (dismissed):
If we want to unify upload/download let's do it in separate PR

NetworkingSampleApp/NetworkingSampleApp/Scenes/Users/UsersViewModel.swift
gajddo00: @matejmolnar I think that was my first comment and conclusion on the Upload PR, that it could have been almost 1:1 with the download manager.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Users/UsersViewModel.swift
gajddo00: No difference in performance between those two btw as long as you don't wish to keep the original capacity.

gajddo00 (dismissed):
👍 Looks good to me

gajddo00 (approved):

cejanen (approved):

Issue #55, opened 2023-12-11
URL encoding and the `+` sign


DanielCech:
I've found one issue in the Networking library. Actually, it is arguable whether it is bug or "feature".
I needed to implement the endpoint with the URL date in ISO8601 format. The problem is the plus sign which is not encoded. On one hand it is correct because plus is URL valid character. On the other hand plus is often interpreted as the space character. And it breaks the communication with BE.

On Lea project I used this simple workaround - I modified the `percentEncodedQuery` of URLComponents.

```swift
func asRequest() throws -> URLRequest {
        var components = try urlComponents()
        components.percentEncodedQuery = components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B") ?? "" // <- Here

        guard let url = components.url else {
            throw RequestableError.invalidURLComponents
        }

        // request setup
        var request = URLRequest(url: url)
        request.httpMethod = method.rawValue
        request.allHTTPHeaderFields = headers
        request.httpBody = try encodeBody()

        // content type
        switch dataType {
        case .encodable:
            request.setValue(
                HTTPHeader.ContentTypeValue.json.rawValue,
                forHTTPHeaderField: HTTPHeader.HeaderField.contentType.rawValue
            )
        case let .custom(_, contentType):
            request.setValue(
                contentType,
                forHTTPHeaderField: HTTPHeader.HeaderField.contentType.rawValue
            )
        default:
            break
        }

        return request
    }
```

Links:
* https://swiftunwrap.com/article/url-components-encoding/
* https://stackoverflow.com/questions/43052657/encode-using-urlcomponents-in-swift

@cejanen knows the details.

PR #56, opened 2023-12-12
[feat] Add support for custom URL parameters percent encoding

matejmolnar:
This PR is trying to solve issue #55

## Changes
Introduce a new type `PercentEncodedParameter` which can be passed as `Any` in `urlParameters: [String: Any]`. The value of this parameter won't get encoded during the `URLRequest` creation.


Sources/Networking/Core/RequestURLParametersType.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  RequestURLParametersType.swift
+//
+//
+//  Created by Matej Molnár on 11.12.2023.
+//
+
+import Foundation
+
+// MARK: - Defines various parameter types to be sent in the URL
```
johnkodes, line 10: Why are we using MARKs on top of the file?
My understanding of MARKs is to use them to split the file into logical parts but I noticed that we use it like this (eg. top of the file) in other places as well.

Is there a reason for that that I'm missing?

Sources/Networking/Core/Requestable+Convenience.swift
```diff
@@ -55,8 +55,13 @@ public extension Requestable {
         }
 
         // encode url parameters
-        if let urlParameters {
-            urlComponents.queryItems = buildQueryItems(urlParameters: urlParameters)
+        if let urlParametersType {
+            switch urlParametersType {
+            case .dictionary(let dictionary):
```
johnkodes, line 60:
```suggestion
            case let .dictionary(dictionary):
```

Sources/Networking/Core/Requestable+Convenience.swift
```diff
@@ -55,8 +55,13 @@ public extension Requestable {
...
+                urlComponents.queryItems = buildQueryItems(urlParameters: dictionary)
+            case .percentEncodedQueryItems(let items):
```
johnkodes, line 62:
```suggestion
            case let .percentEncodedQueryItems(items):
```

johnkodes (dismissed):

Sources/Networking/Core/RequestURLParametersType.swift
```diff
@@ -0,0 +1,20 @@
...
+
+/// A parameter type of request URL.
+public enum RequestURLParametersType {
+    /// A dictionary which will get transformed into an array of `URLQueryItem` and then it will get automatically percent-encoded by `URLComponents` before the request.
+    case dictionary([String: Any])
+    /// An array of already percent-encoded URL query items.
+    ///
+    /// This can be useful in case the default percent-encoding doesn't suit your use-case. Be aware that it assumes the query item names and values are already correctly percent-encoded, and that the query item names do not contain the query item delimiter characters '&' and '='. Attempting to set an incorrectly percent-encoded query item or a query item name with the query item delimiter characters '&' and '=' will cause a `fatalError`.
```
johnkodes, line 18: Please split this into multiple lines

Sources/Networking/Core/RequestURLParametersType.swift
matejmolnar: Agreed, I don't see good reason for the MARKs either. Shall I create another PR to remove these unnecessary MARKs?

johnkodes: Do you know who started this convention? 😃 As I don't want to overrule that without first asking, but otherwise I would be for this in a new PR 😊.

matejmolnar: Not sure, maybe @cejanen?

cejanen:
@matejmolnar I played with it for a bit and I would propose a bit difference solution. Custom string encoding similar like specific array encoding is edge case. In 99% cases user is ok with parameters [String: Any]. Therefore I would not push them to wrap always parameters into `RequestURLParametersType`. I would better go with similar approach like we have for ArrayParameter. 
We can do CustomEncodingParameter with either string which will be encoded for URL automatically or allow already encoded values.
Code which may help (draft):
`if let percentedEncoded = value as? PercentEncodedParameter {
            queryItems.append(
                URLQueryItem(
                    name: key,
                    value: percentedEncoded.value
                )
                .percentEncodedWithValue(string: percentedEncoded.value)
            )
            return queryItems
        }`

gajddo00:
@cejanen Agree, the intention behind designing ArrayParameter as it is was to avoid publishing a new breaking change to the library and making it optional for the user at the same time. For context, this is the `Requestable+Convenience` query building implementation, I'm also adding an adjusted sample from CJs draft if it makes sense.

```swift
// MARK: Build Query Items
private extension Requestable {
  func buildQueryItems(urlParameters: [String: Any]) -> [URLQueryItem] { ... }

  func buildQueryItems(key: String, value: Any) -> [URLQueryItem] {
      switch value {
      case let parameter as ArrayParameter:
          return buildArrayParameter(key: key, parameter: parameter)
       case let parameter as PercentEncodedParameter:
           return buildPercentEncodedParameter(key: key, parameter: parameter)
       default:
           return [URLQueryItem(name: key, value: String(describing: value))]
      }
  }
  
  func buildArrayParameter(
        key: String,
        parameter: ArrayParameter
    ) -> [URLQueryItem] { ... }
    
    func buildPercentEncodedParameter(
        key: String,
        parameter: PercentEncodedParameter
    ) -> [URLQueryItem] {
        [URLQueryItem(name: key, value: parameter.percentEncodedValue)]
    }
}

public struct PercentEncodedParameter {
    let value: String

    public var percentEncodedValue: String? {
        value.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)?
            .replacingOccurrences(of: "+", with: "%2B")
    }

    public init(value: String) {
        self.value = value
    }
}
```

matejmolnar:
@cejanen @gajddo00 Thank you for the suggestions 🙏, I agree that it's a better approach although I had to adjust it a little.

gajddo00 (dismissed):
Looks good to me 👍

cejanen (changes requested):
I tested it and it doesn't work as expected. Will try to do updates. It wasn't very clear from my side.

tested URL params
page=1&test=1+1

NetworkingSampleApp/NetworkingSampleApp/API/Routers/SampleUserRouter.swift
```diff
@@ -30,11 +30,11 @@ enum SampleUserRouter: Requestable {
...
         }
     }
-
-    var urlParameters: [String: Any]? {
+    
+    var urlParametersType: RequestURLParametersType? {
         switch self {
         case let .users(page):
-            ["page": page]
+            .dictionary(["page": page])
```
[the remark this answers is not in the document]
matejmolnar: I think that creating a few unit tests would be better for this purpose.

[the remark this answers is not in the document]
cejanen: We need to add them but first I will fix the PR, right now it doesn't work as expected. Testing why...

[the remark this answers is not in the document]
matejmolnar: I've added a few tests in `URLParametersTests` to demonstrate how I understood the issue that we are trying to solve. Am I missing something?

cejanen (approved):

cejanen:
@gajddo00 @matejmolnar At the end I decided for simplest solution. By default are data encoded on URLComponent and you can pass custom encoded data which are expected like encoded already. All tests pass multiple times.

Sources/Networking/Utils/String+PlusSignEncoded.swift
```diff
@@ -0,0 +1,17 @@
...
+import Foundation
+
+public extension String {
+    /// Help method to allow custom + sign encoding, more in ```CustomEncodedParameter```
+    func plusSignEncoded() -> Self? {
+        self
+            .addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)?
+            .replacingOccurrences(of: "+", with: "%2B")
+    }
+}
```
matejmolnar, line 17: I think we could just move this extension to the URLParametersTests file, since it's not being used anywhere else 🤔

Sources/Networking/Misc/CustomEncodedParameter.swift
```diff
@@ -0,0 +1,35 @@
...
+/// Usage example:
+///
+///     var urlParameters: [String: Any]? {
+///         ["specialCharacter": ">"]
+///     }
+///
+///     // Request URL "https://test.com?specialCharacter=%3E"
+///
+///     var urlParameters: [String: Any]? {
+///         ["specialCharacter": PercentEncodedParameter(">")]
```
matejmolnar, line 23:
```suggestion
///         ["specialCharacter": CustomEncodedParameter(">")]
```

gajddo00 (approved):
🍓 Agreed as in most cases custom encoding won't be necessary, we can elevate this to the user, why not.

Issue #42, opened 2023-06-10
README


johnkodes:
Let's create a typical library README

Some inspiration:
- https://github.com/kean/Get 
- https://github.com/strvcom/ios-dependency-injection
- https://github.com/Alamofire/Alamofire/blob/master/Documentation/Usage.md#introduction

PR #57, opened 2023-12-13
[feat] Add proper README

matejmolnar:
This PR aims to solve issue #42 

## Changes
- Updated README with a structure similar to https://github.com/strvcom/ios-dependency-injection/tree/master
- Updated Documentation with UploadAPIManager 

## Discussion
- I am not sure whether the README is not too long and it wouldn't be better to leave some stuff out.
- Should I add Roadmap at the end similarly to Dependency Injection library?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,323 @@
 # Networking
-The streamlined library for efficient API call management. This lightweight solution leverages the power of Swift concurrency by building upon URL sessions.
 
-The library is thoughtfully documented using the DocC documentation format, ensuring comprehensive and accessible documentation for developers.
+[![Coverage](https://img.shields.io/badge/Coverage-100%25-darkgreen?style=flat-square)]
+[![Platforms](https://img.shields.io/badge/Platforms-iOS_iPadOS_macOS_watchOS-lightgrey?style=flat-square)]
+[![Swift](https://img.shields.io/badge/Swift-5.9+-blue?style=flat-square)]
 
-## Supported features
-TBD
+A networking layer using native `UrlSession` and Swift concurrency.
+
+## Requirements
+
+- iOS/iPadOS 15.0+, macOS 12.0+, watchOS 9.0+
+- Xcode 14+
+- Swift 5.9+
+
+## Installation
+
+You can install the library with [Swift Package Manager](https://swift.org/package-manager/). Once you have your Swift package set up, adding Dependency Injection as a dependency is as easy as adding it to the `dependencies` value of your `Package.swift`.
+
+```swift
+dependencies: [
+    .package(url: "https://github.com/strvcom/ios-networking.git", .upToNextMajor(from: "0.0.4"))
+]
+```
+
+## Overview
+Heavily inspired by Moya, the networking layer's philosophy is focused on creating individual endpoint routers, transforming them into a valid URLRequest objects and applying optional adapters and processors in the network call pipeline.
+
+## Router
+By conforming to the ``Requestable`` protocol, you can define endpoint definitions containing the elementary HTTP request components necessary to create valid HTTP requests.
+<br>**Recommendation:** Follow the `Router` naming convention to explicitly indicate the usage of a router pattern.
+
+### Example
+```swift
+enum UserRouter { 
+    case getUser
+    case updateUser(UpdateUserRequest)
+}
+
+extension UserRouter: Requestable {
+    // The base URL address used for the HTTP call.
+    var baseURL: URL { 
+        URL(string: Constants.baseHost)!
+    }
+
+    // Path will be appended to the base URL.
+    var path: String { 
+        switch self {
+        case .getUser, .updateUser:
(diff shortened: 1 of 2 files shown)


johnkodes:
> I am not sure whether the README is not too long and it wouldn't be better to leave some stuff out.

@matejmolnar I think the length is fine 👍 (we can probably also add something like a "table of contents" in the Overview section. In general this is a great documentation but I had a little bit different structure of README in mind when I was writing the Github issue. 

To me the README should consist and describe use cases that I can achieve with the library. When I come to a library and I know nothing about it, reading about each object isn't what I'm searching for. I'm searching for: "Making a simple request:", "Downloading a file", "Uploading data" (it can be something different, but you get the idea). And inside those sections I can learn briefly about the `DownloadAPIManager` and what that we use Routers for example with links to either another section in the README for this specific subject that would just talk about those things in detail (as you already kinda have) or possibly a DocC documentation (we have that right? It's just not uploaded anywhere so that we could browse that in a web browser)?

Then we can have a section like "Advanced Usage" that would describe some more advanced topics like the Authorization, Adapters (if we consider those advanced) and so on. We can definitely discuss the individual sections before if we decide to change it into this. 

But this is just my personal opinion. What do you think about it? @matejmolnar And what do others @strvcom/ios-code-owners think?

matejmolnar:
@johnkodes Makes sense, I will try to rework it.
Yeah we have DocC documentation and it's not uploaded anywhere as far as I know. Is it customary to do so? 🤔

matejmolnar:
@johnkodes I've updated the README. let me know what you think.

README.md
```diff
@@ -1,18 +1,39 @@
...
+- [Modifiers](#modifiers)
+- [Interceptors](#interceptors)
+    - [LoggingInterceptor](#loggingInterceptor)
+    - [AuthorizationTokenInterceptor](#authorizationTokenInterceptor)
+- [Processors](#processors)
+    - [StatusCodeProcessor](#statusCodeProcessor)
+    - [EndpointRequestStorageProcessor](#endpointRequestStorageProcessor)
+- [Associated array query parameters](#arrayQueryParameters)
+
+## Requirements <a name="requirements"></a>
```
johnkodes, line 30: Doesn't it work automatically without doing those anchors? 🤔 Because it should 😄

matejmolnar: Good point, done ✅

cejanen (commented):
Screenshot 2024-01-08 at 21 29 17

I haven't read it all yet. I found this broken formatting. Overall I think the readme format can have more free spaces to increase readability. Thinking about classes description - maybe we can remove those from readme and add it as a part of docC document - that means we don't need to maintain it twice. What do you think?

johnkodes (approved):
@matejmolnar  Thank you for the changes 👍. It looks much better now. I have one last ask about that and that is if you could run it through Grammarly and I also use a markdown formatter. I'm using [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) for that inside VS Code. 

Second note comes from a discussion with @cejanen . CJ mentioned that the README might be too long and we might not need all the descriptions of the actual objects. But it wouldn't mean that we should trash everything you created. We would like to explore DocC that would be hosted on Github. Similar how other libraries do it (https://pointfreeco.github.io/swift-composable-architecture/main/documentation/composablearchitecture/store). We hope it will be possible under our STRV organization but this would be on you to research, to try to figure out how to generate + deploy. The idea would be most likely to shorten the README as it is now and for details link to this documentation where we would move some parts of this README. We already have some Docc but it seems outdated compared to this README, so we would update it. And additionally to the generated Docc from the codebase you can also add whatever you want as another markdowns inside it, so we could leverage that to add more detailed information about some stuff as well if necessary. I suppose this would be in another PR and for the time being we would merge this PR and if we replace some parts we would adjust that later. 

What do you think about that? ☺️

cejanen:
Follow up to @johnkodes comment. This may be helpful generating DocC as a web page on github https://apple.github.io/swift-docc-plugin/documentation/swiftdoccplugin/publishing-to-github-pages/. In readme I would keep usage, installation, mention some more specific features like array encoding option but classes description etc I would move to DocC and only add link to documentation into readme. 
With that comes in mind one extra subtask. We need refresh texts in DocC as well. We may check comments in classes and we can eventually remove MARKs at the top of classes or replace with DocC supported comments. 

It may look extremely heavy but I don't think so. We have good readme only format can be nicer. From readme we can take text for docC - try to generate it in project and update texts in docC structure and classes as well. 

Still whole PR is about the texts, we don't touch code, not a big deal if we make "mistake" here.

cejanen:
Noticed @johnkodes to do the rest of work in another PR - agree. There fore approving and @matejmolnar you can take our last comments as a start for another PR 
🙇‍♂️

cejanen (approved):
Good progress with documentation. Let's continue in another PR

PR #59, opened 2023-12-21
[feat] Downloads/Uploads manager usage simplification and sample unification

matejmolnar:
This PR aims to simply the usage of DownloadAPIManager/UploadAPIManager and also unify the sample screens logic + UI.

## Managers usage simplification
While I was writing a new README I realised that we are forcing the user to create unnecessary Router boiler plate in order to create a download/upload request. 
### Download
Hence I've created a new method on `DownloadAPIManaging` where you can simply pass the `URL` of the file you wish to download instead of `Requestable`. The request is then made with internal dummy `DownloadRouter`.
```.swift
    func downloadRequest(
        _ fileURL: URL,
        resumableData: Data? = nil,
        retryConfiguration: RetryConfiguration? = .default
    ) async throws -> DownloadResult {
        try await downloadRequest(
            DownloadRouter(fileURL: fileURL),
            resumableData: resumableData,
            retryConfiguration: retryConfiguration
        )
    }
```
### Upload
First I've created a new enum `UploadType` and I replaced the three different upload methods with one where you pass an uploadType parameter. Then I defined a new method where you can pass the `UploadType` and an upload `URL`. Similarly to download the request is made with internal dummy `UploadRouter`.

```.swift
    func upload(_ type: UploadType, to uploadURL: URL) async throws -> UploadTask {
        try await upload(type, to: UploadRouter(url: uploadURL, uploadType: type))
    }
```

## Sample logic + UI unification
- Create `UploadAPIManager` shared instance.
- Remove `UploadService` and use the shared manager instance instead.
- Remove `FormUploadsViewModel` and move the logic to the `UploadsViewModel`.
- Create new `TaskProgressView` with generic viewModel of `TaskProgressViewModel` protocol and use it in both upload and download screens.


NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressViewModel.swift
```diff
@@ -9,28 +9,43 @@ import Foundation
...
 
 @MainActor
-final class DownloadProgressViewModel: ObservableObject {
+final class DownloadProgressViewModel: TaskProgressViewModel {
     private let task: URLSessionTask
     
-    @Published var state: DownloadProgressState = .init()
-    
+    let isRetryable = false
+    private(set) var title: String = ""
```
gajddo00, line 16: Why not have a @Published private(set) var state: DownloadState instead?

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadsView.swift
```diff
@@ -53,14 +39,9 @@ struct UploadsView: View {
...
-        .alert(
-            "Error",
-            isPresented: $formViewModel.isErrorAlertPresented,
-            actions: {},
-            message: {
-                Text(formViewModel.error?.localizedDescription ?? "")
-            }
-        )
+        .onAppear {
+            viewModel.loadTasks()
```
gajddo00, line 43: We could use .task { } here and get auto cancel as loadTasks() does async work anyways.

NetworkingSampleApp/NetworkingSampleApp/Scenes/Download/DownloadProgressViewModel.swift
matejmolnar: I reworked it to conform to the `TaskProgressViewModel` protocol.

Sources/Networking/Core/Upload/UploadType.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  File.swift
```
cejanen, line 2: I found wrong filename + swiftlint warning here

Sources/Networking/Core/DownloadAPIManaging.swift
```diff
@@ -41,13 +41,52 @@ public protocol DownloadAPIManaging {
...
-        try await downloadRequest(endpoint, resumableData: resumableData, retryConfiguration: retryConfiguration)
+        try await downloadRequest(
+            endpoint,
+            resumableData: resumableData,
+            retryConfiguration: retryConfiguration
+        )
+    }
+}
+
+/// A Router used for basic use case of downloading a file from a URL.
```
cejanen, line 77: I'm not sure if I like having this private struct in the same file as is the public interface to the library 🤔 . Wouldn't it be better somewhere else (own file?)

Sources/Networking/Core/Upload/UploadAPIManaging.swift
```diff
@@ -73,29 +45,46 @@ public protocol UploadAPIManaging {
...
-    
+
     /// Returns an active ``UploadTask`` specified by its identifier.
     func task(with id: UploadTask.ID) async -> UploadTask? {
         await activeTasks.first { $0.id == id }
     }
 }
+
+/// A Router for basic use case of uploading file/data/multiPartForm to a given URL.
+private struct UploadRouter: Requestable {
```
cejanen, line 64: I have similar feeling here as in the downloading managing (private struct in public interface)

cejanen (changes requested):
We're almost there. I think about few little improvements we can do here.

NetworkingSampleApp/NetworkingSampleApp/UIElements/TaskProgressView.swift
```diff
@@ -0,0 +1,70 @@
...
+    var status: String { get }
+    var downloadedBytes: String { get }
+    var state: URLSessionTask.State { get }
+    var percentCompleted: Double { get }
+    var isRetryable: Bool { get }
+
+    func suspend()
+    func resume()
+    func cancel()
+    func retry()
```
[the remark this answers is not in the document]
matejmolnar: I don't really see a benefit to do that in this case. Do you have anything specific in mind?

NetworkingSampleApp/NetworkingSampleApp/Scenes/Upload/UploadsViewModel.swift
```diff
@@ -7,33 +7,40 @@
 
 import Foundation
 import OSLog
+import Networking
```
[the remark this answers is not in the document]
matejmolnar: I have never though of that actually 😅

Sources/Networking/Core/DownloadAPIManaging.swift
matejmolnar: Sure I can do that if you want. I just like to put simple single use types into the source file where they are used in, since they don't need to be exposed 🤷‍♂️

cejanen (dismissed):
🩴

cejanen (approved):
it took me only ages but at the end I'm happy with result 
📲

gajddo00 (approved):
🐱

PR #60, opened 2024-01-11
[feat] Make Authorization data codable

raulbatista94:
Since we are planning to use this model in the template and store it to keychain the current implementation of Keychain manager in the template requires the model to conform Codable protocol.


matejmolnar (approved):
Makes sense

matejmolnar (approved):

cejanen (approved):
🦖

PR #61, opened 2024-01-12
Feat/support ios14

cejanen:
There was a request to support version of iOS. This PR changes supported iOS version and fix all necessary places when iOS 15 is required. 
In common the iOS 14 support is without upload task, for iOS 14 is generated sessionID via formatter instead of iOS15 function formattedDate. 
Tests passed.


Sources/Networking/Core/APIManager.swift
```diff
@@ -37,7 +44,13 @@ open class APIManager: APIManaging, Retryable {
         errorProcessors: [ErrorProcessing] = []
     ) {
         /// generate session id in readable format
-        sessionId = Date().ISO8601Format()
+        if #unavailable(iOS 15) {
```
gajddo00, line 47: Maybe this could be reused? I see it three times with the same exact code.

gajddo00 (dismissed):
🛸 Looks good, bummer that this PR contains few other branches as well. I assume those were already reviewed and it's enough to check the last iOS 14 support commit?

Sources/Networking/Core/DownloadAPIManager.swift
```diff
@@ -34,7 +34,13 @@ open class DownloadAPIManager: NSObject, Retryable {
         errorProcessors: [ErrorProcessing] = []
     ) {
         /// generate session id in readable format
-        sessionId = Date().ISO8601Format()
+        if #unavailable(iOS 15) {
```
ipek, line 37: Why is the if sometimes `#available` and sometimes `#unavailable` if they both have `else`?

Sources/Networking/Core/Upload/MultipartFormDataEncoder.swift
```diff
@@ -31,14 +31,9 @@ open class MultipartFormDataEncoder {
...
 
-
-/**
- 
-The main reason why there are methods to encode data & encode file is similar to `uploadTask(with:from:)` and `uploadTask(with:fromFile:)`  ig one could convert the content of the file to Data using Data(contentsOf:) and use the first method to send data. One has the data available in memory while the second reads the data directly from the file thus doesn't load the data into memory so it is more efficient.
- */
-
 // MARK: - MultipartFormDataEncoding
 extension MultipartFormDataEncoder: MultipartFormDataEncoding {
+    /// The main reason why there are methods to encode data & encode file is similar to `uploadTask(with:from:)` and `uploadTask(with:fromFile:)`  ig one could convert the content of the file to Data using Data(contentsOf:) and use the first method to send data. One has the data available in memory while the second reads the data directly from the file thus doesn't load the data into memory so it is more efficient.
```
ipek, line 36:
```suggestion
    /// The main reason why there are methods to encode data & encode file is similar to `uploadTask(with:from:)` and `uploadTask(with:fromFile:)` if one could convert the content of the file to Data using Data(contentsOf:) and use the first method to send data. One has the data available in memory while the second reads the data directly from the file thus doesn't load the data into memory so it is more efficient.
```

ipek (dismissed):
Looks good. Just one thought and noticed one typo.

Sources/Networking/Core/APIManager.swift
cejanen: yes but original it is one line command. It will be back soon I hope

Sources/Networking/Core/DownloadAPIManager.swift
cejanen: Just for feeling of readability. Available I used to hide feature unavailable to see immediately there is fall back to old versions

PR #62, opened 2024-01-17
[feat]: Add CI workflow to run tests

johnkodes:
## Changes: 
- Adds a workflow that should run tests on different versions of iOS + MacOS and also with different Xcodes. 
  - Triggered on PRs and pushes to `master`
 - Adds test results via https://github.com/kishikawakatsumi/xcresulttool
   - only for iOS because macOS uses `swift test` instead of `xcodebuild` which is faster twice faster

## Limitations:
- We cannot test lower iOS versions because we build using Swift 5.9 (which requires Xcode 15) and the Github Runners don't have older simulators installed. https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode
- We cannot test the macOS version via `swift test` because the loading of files fails (resources are handled differently when running `swift test` vs `xcodebuild`). When running `swift test` they are not compiled into `Assets.car` which might be the reason why we are unable to load the file via `NSDataAsset` (in `MockResponseProvider`) but I haven't debugged that far to know that for sure.


cejanen (approved):
I don't see any issue we run on latest iOS & xcode etc. The main goal is to build library spm, sample app performance we can check manually from time to time on older iOSes. 
Best test is that we use library in our project and it works :-D

Sources/Networking/Modifiers/Processors/EndpointRequestStorageProcessor/EndpointRequestStorageProcessor.swift
```diff
@@ -137,8 +137,8 @@ private extension EndpointRequestStorageProcessor {
         _ response: Response,
         endpointRequest: EndpointRequest,
         urlRequest: URLRequest
-    ) {
-        Task(priority: .background) { [weak self] in
+    ) async {
+        await Task(priority: .background) { [weak self] in
```
johnkodes, line 141: Please check those changes @cejanen @matejmolnar

Sources/Networking/Modifiers/Processors/EndpointRequestStorageProcessor/EndpointRequestStorageProcessor.swift
```diff
@@ -87,9 +87,9 @@ open class EndpointRequestStorageProcessor: ResponseProcessing, ErrorProcessing
         
         switch error {
         case let .unacceptableStatusCode(_, _, response):
-            storeResponse(response, endpointRequest: endpointRequest, urlRequest: urlRequest)
+            await storeResponse(response, endpointRequest: endpointRequest, urlRequest: urlRequest)
```
johnkodes, line 90: Please check those changes @cejanen @matejmolnar

Tests/NetworkingTests/Mocks/MockResponseProviderTests.swift
cejanen, line 1: I believe this is a misunderstanding. MockResponseProvider is valid class of library. Those are test for that class, not creating mock class.
MockResponseProvider - is open class in library. We may rename the class to StoredResponseProvider to avoid confusion

cejanen (changes requested):

johnkodes:
> I don't see any issue we run on latest iOS & xcode etc. The main goal is to build library spm, sample app performance we can check manually from time to time on older iOSes. 
> 
> Best test is that we use library in our project and it works :-D 

@cejanen SPM builds are usually tested using `swift test` but unfortunately those fail because of the implementation of the mocked data loader as described in the PR description. 

We can run `swift build` to verify that the package builds without errors.

Or did you mean something else?

Tests/NetworkingTests/Mocks/MockResponseProviderTests.swift
johnkodes: My bad, I'll move the file back. Plus also I shouldnt move it here cause it's a test 🫣

Tests/NetworkingTests/Mocks/MockFileManager.swift
```diff
@@ -0,0 +1,49 @@
...
+//
+//  Created by Jan Kodeš on 23.01.2024.
+//
+
+import Foundation
+import Foundation
+import XCTest
+
+/// A subclass of `FileManager` where the file existence is based on a dictionary whose key is the file path.
+final class MockFileManager: FileManager {
```
matejmolnar, line 13: Good idea 👍

Sources/Networking/Modifiers/Processors/EndpointRequestStorageProcessor/EndpointRequestStorageProcessor.swift
matejmolnar: I don't think we want to do this. It was specifically written to execute in a background task in order to not delay requests processing in the app. Which actually reminds me that the Task should also be detached.

johnkodes: Quick question before I try different solutions how to deterministically write tests without using the sleeps. 

Was this really an issue? That the Filemanager/Encoder was slow? 

And we are talking about debug/test builds only right? This isnt supposed to be used in production.

cejanen: 1) this processor should be used only in debug 
2) even in debug we didn't want to block response processing just by storing the response to have mock data available, therefore the task was not awaiting
3) I don't like the sleep as well but if I remember correctly if the test was run immediately the file was not stored yet (some delay there is)

johnkodes: @cejanen 3. not sure if understand this point entirely 😅 . The reason why I did those changes is to know when this `Task` finished (so that I know in the test), because it is run on a BG thread.

matejmolnar: I am not aware of a way to test the background task write operations on FileManger deterministically. But I suppose we could add some observation logic to the newly added MockFileManager 🤔

johnkodes: Done 👍, please have a look. The `FileManager` can't observe files and it probably wouldn't help because the final data writing was done via `Data.write`.

Tests/NetworkingTests/EndpointRequestStorageProcessorTests.swift
```diff
@@ -116,19 +130,17 @@ final class EndpointRequestStorageProcessorTests: XCTestCase {
...
-            model.method == "GET" &&
-            model.path == mockEndpointRequest.endpoint.path &&
-            model.parameters == ["query": "mock"] &&
-            model.requestBody == nil &&
-            model.requestBodyString == nil &&
-            model.requestHeaders == mockURLRequest.allHTTPHeaderFields &&
-            model.responseBody == mockResponseData &&
-            model.responseBodyString == String(data: mockResponseData, encoding: .utf8) &&
-            model.responseHeaders == ["mockResponseHeader": "mock"]
-        )
```
johnkodes, line 131: @matejmolnar please don't use something like this. It makes it very hard to understand what is wrong when the test fails.

matejmolnar: Good point 👍, thanks for the feedback.

Sources/Networking/Utils/FileDataWriter.swift
```diff
@@ -0,0 +1,26 @@
+//
+//  FileWriter.swift
+//
+//
+//  Created by Jan Kodeš on 24.01.2024.
+//
+
+import Foundation
+
+public protocol FileDataWriterProtocol {
```
matejmolnar, line 10: Wouldn't a `FileDataWriting` be a better name since we are using the `ing` convention throughout the project? Also could you please include at least one sentence for both the protocol and the `FileDataWriter` for documentation? 🙏

Tests/NetworkingTests/Mocks/FileWriterSpy.swift
```diff
@@ -0,0 +1,25 @@
...
+//  FileDataWriterSpy.swift
+//
+//
+//  Created by Jan Kodeš on 24.01.2024.
+//
+
+import Foundation
+import Networking
+
+class FileDataWriterSpy: FileDataWriterProtocol {
```
matejmolnar, line 11: Could you please include at least one sentence about the object for documentation? 🙏

matejmolnar (changes requested):

Sources/Networking/Modifiers/Processors/EndpointRequestStorageProcessor/EndpointRequestStorageProcessor.swift
matejmolnar: Oh right, I didn't realise that 😅 I like the new implementation 🙂

Sources/Networking/Utils/FileDataWriter.swift
johnkodes: Definitely!

Sources/Networking/Modifiers/Processors/EndpointRequestStorageProcessor/EndpointRequestStorageProcessor.swift
cejanen: > @cejanen 3. not sure if understand this point entirely 😅 . The reason why I did those changes is to know when this `Task` finished (so that I know in the test), because it is run on a BG thread.

I understand why you did it. Only saying we don't want it :-D For test convenience you changed the behaviour of that feature which is inconvenient 🦜

Sources/Networking/Modifiers/Processors/EndpointRequestStorageProcessor/EndpointRequestStorageProcessor.swift
```diff
@@ -138,7 +141,7 @@ private extension EndpointRequestStorageProcessor {
         endpointRequest: EndpointRequest,
         urlRequest: URLRequest
     ) {
-        Task(priority: .background) { [weak self] in
+        Task.detached(priority: .utility) { [weak self] in
```
cejanen, line 144: Although I think logical to have it detached, I'm just wondering if it would work also without having it detached?

Tests/NetworkingTests/Mocks/MockFileDataWriter.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+//  Created by Jan Kodeš on 24.01.2024.
+//
+
+import Foundation
+import Networking
+
+/// A test mock class for `FileDataWriting`.
+/// It writes into a file but let's us react when it's finished.
+class MockFileDataWriter: FileDataWriting {
```
cejanen, line 13: We have MockFileManager final (I assume all mock classes) can we have it final as well?

Tests/NetworkingTests/Mocks/MockFileDataWriter.swift
```diff
@@ -0,0 +1,27 @@
...
+    var writeClosure: (() -> Void)?
+    private(set) var writeCalled = false
```
cejanen, line 15: why do we have properties writeCalled, receivedData or receivedURL when we don't use them anywhere or anyway?

cejanen (changes requested):
Leaving few more questions/comments Ω

Sources/Networking/Modifiers/Processors/EndpointRequestStorageProcessor/EndpointRequestStorageProcessor.swift
johnkodes: Yes, it would work. This was @matejmolnar 's idea, which made sense to me, so I added that.

Tests/NetworkingTests/Mocks/MockFileDataWriter.swift
johnkodes: It's a practice that I have when writing any mocks, so that if needed anyone can use those properties to verify that it was called properly in the tests.

Sources/Networking/Modifiers/Processors/EndpointRequestStorageProcessor/EndpointRequestStorageProcessor.swift
johnkodes: @cejanen yep, got it. Reworked 👍

Tests/NetworkingTests/Mocks/MockFileDataWriter.swift
cejanen: When we have it can we also use it in tests? Aka write a test which  it uses. 

>  if needed anyone can use those properties to verify that it was called properly in the tests
I can write tons of code that anyone can potentially use in future, but usually until people know it they create their own solution anyway 🍶

matejmolnar (dismissed):
📝

johnkodes:
@matejmolnar @cejanen I hope you noticed but just to be sure that you've seen those as well 😄 
- https://github.com/strvcom/ios-networking/pull/62/checks?check_run_id=20976997060&pr=62
- https://github.com/strvcom/ios-networking/pull/62/checks?check_run_id=20976737145&pr=62

Found the **checks**:
CleanShot 2024-01-31 at 11 51 01

Tests/NetworkingTests/Mocks/MockFileDataWriter.swift
johnkodes: Added. I refactored the tests a bit and used the `MockFileWriter`.

johnkodes:
@cejanen @matejmolnar Few more changes added in the tests. Let's have a look again please. Possibly the last review 😄

cejanen (approved):
👍

matejmolnar (approved):
💪

PR #63, opened 2024-01-23
[feat] Web hosted documentation

matejmolnar:
This PR aims to solve this [ticket](https://strvcom.atlassian.net/jira/software/projects/IOSIP/boards/308?selectedIssue=IOSIP-24)

- Remove unnecessary pragma marks
- Add swift setting flag to enable DocC symbol links for code extensions
- Add swift DocC plugin and `generate_docs.sh` script file for generating static web documentation files
- Move classes related documentation to appropriate files instead of keeping it in `documentation.md`/`README.md`
- Update outdated documentation
- Update README with links to the web hosted documentation, which won't be working until we push the changes to master and someone with admin access sets the github pages like so:
Screenshot 2024-01-17 at 15 08 39


cejanen:
Just sharing first feelings... 793 files changed easily...  🧩

matejmolnar:
@cejanen I know 😅, but It's mostly generated files for the static web docs.

cejanen (commented):
Honestly I haven't read all texts but "smekam". I don't know if you used any special tool to generate those texts specially to showcase usage of classes but respect. 
So far I'm just curious why do we remove readme file? 
Another question. Can I test somehow that the docs will work as expected before merging and updating repo settings as described in PR notes.

README.md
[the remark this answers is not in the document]
matejmolnar: The readme is still there, I've just updated it. Not sure why github thinks it's removed 🤔

matejmolnar:
> Honestly I haven't read all texts but "smekam". I don't know if you used any special tool to generate those texts specially to showcase usage of classes but respect. So far I'm just curious why do we remove readme file? Another question. Can I test somehow that the docs will work as expected before merging and updating repo settings as described in PR notes.
Thank you for the kind words 🙏. No I didn't use any tool to generate the usage documentation.

You can test the documentation by running
```
swift package --disable-sandbox preview-documentation --target Networking
```
It will start a local preview server with the documentation.

johnkodes:
@matejmolnar `someone with admin access sets the github pages like so:` this is done 👍

cejanen (approved):
Wanna see it live 🥇

Package.swift
```diff
@@ -17,12 +17,16 @@ let package = Package(
...
+    dependencies: [
+        .package(url: "https://github.com/realm/SwiftLint.git", exact: "0.53.0"),
+        .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0")
+    ],
     targets: [
         // Targets are the basic building blocks of a package. A target can define a module or a test suite.
         // Targets can depend on other targets in this package, and on products in packages this package depends on.
         .target(
             name: "Networking",
+            swiftSettings: [.unsafeFlags(["-emit-extension-block-symbols"])],
```
johnkodes, line 29: Could you forward somewhere where I can read more about it? 😄

Sources/Networking/Core/APIManager.swift
```diff
@@ -7,7 +7,37 @@
 
 import Foundation
 
-/// Default API manager
+/** Default API manager which is responsible for the creation and management of network requests.
```
johnkodes, line 10: Why are we using the `/**` here instead of staying with `///`?

docs/css/chunk-c0335d80.10a2f091.css
johnkodes, line 1: How come that for example TCA doesn't have all of those CSS/JS files 🤔 😄 ? Do you know how can that be achieved? 

https://github.com/pointfreeco/swift-composable-architecture/tree/main/Sources/ComposableArchitecture/Documentation.docc

Package.swift
matejmolnar: Yeah so the problem was, that I wasn't able to reference extensions of types which are not defined in this module. So for example when I wanted to reference ``URLSessionTask.DownloadState`` it would not create the link. I found the solution here https://github.com/apple/swift-docc/issues/210#issuecomment-1343975239

Sources/Networking/Core/APIManager.swift
matejmolnar: I've used it only for longer descriptions, especially with code examples etc. It's just easier to edit and keep the proper spacing that way.

docs/css/chunk-c0335d80.10a2f091.css
matejmolnar: Some projects use a special branch which includes just the documentation, so that's what I though they were doing as well. But upon a closer inspection I found out that it's not the case, but they are actually using [Jekyll](https://jekyllrb.com/docs/github-pages/) to build the documentation in a [github action](https://github.com/pointfreeco/swift-composable-architecture/actions/workflows/pages/pages-build-deployment)

Sources/Networking/Core/APIManager.swift
johnkodes: Got it, I would still prefer the `///` everywhere, but we can go with this 👍

docs/css/chunk-c0335d80.10a2f091.css
matejmolnar: Ah good catch, how did you find it? 😄 So yeah that's how they do it and they also use Jekyll on top. When I was talking to @cejanen about the separate branch, he said to not worry about it for now and just put the docs into the master branch. But yeah once we would like to also keep multiple versions of the documentation a separate branch is the way to go 🙂

johnkodes (approved):
I would love if we could have the docs hidden elsewhere (eg. in the separate branch), but this is more than enough ☺️ . Thank you 👍

PR #67, opened 2024-02-29
[CHORE] Use docc plugin param instead of unsafe flag

gajddo00:
https://forums.swift.org/t/enablement-of-docc-extension-support-as-an-opt-in-feature/62614


gajddo00:
@matejmolnar Not sure if you had problem with extended modules in Xcode docc or the generated static files. But they should by now be using the flag by default.

matejmolnar (approved):
Looks good, thank you 🙏

cejanen (approved):

PR #68, opened 2024-04-10
[fix] URL query keys percent encoding

matejmolnar:
Add percent encoding also for keys in url queries. Otherwise adding some non allowed character to the url query key would crash the app 😬


arkhigleb (approved):

cernym46 (approved):

cejanen (approved):
🛋️

PR #69, opened 2024-05-08
Sorted url query parameters for better caching support

gajddo00:
Enhances the possibility for URL addresses to be cacheable.


matejmolnar (approved):
⚽
I would just add that this change is necessary due to how CDNs work. Usually they determine whether to return a cached response for a request simply based on the url string comparison, which means that if you happen to switch the order of url parameters between requests it doesn't return the cached response.

arkhigleb (approved):

PR #70, opened 2024-07-29
fix: rule out watchos directive for mutlipeer connectivity processor

gajddo00:
Multi peer Connectivity is not supported on watchOS.

https://developer.apple.com/documentation/multipeerconnectivity


matejmolnar (approved):
⌚

cejanen (approved):
🐌

PR #71, opened 2024-07-31
[fix] allow to override default json decoder on APIManager

cejanen:
The original solution had default json encoder defined on APIManaging protocol and APIManager had not exposed this public variable so there was not a easy way to override default json decoder in APIManager subclasses.

I removed the default implementation on the protocol to force any APIManaging implementation to expose this variable. 
At the same time I don't want change init method for APIManager - any child class can solve it with custom init or just override the public property


gajddo00 (approved):
🍏

fiebiger (approved):
💯

PR #73, opened 2025-02-03
chore: url session invalidation

gajddo00:
Adds support for URLSession invalidation to `APIManaging`.


Sources/Networking/Core/APIManaging.swift
```diff
@@ -11,10 +11,10 @@ import Foundation
 // MARK: - Defines API managing
 
 /// A definition of an API layer with methods for handling API requests.
-public protocol APIManaging {
+public protocol APIManaging: URLSessionInvalidatable {
```
cejanen, line 14: just a philosophical question. Do we need to APIManaging to conform this or we can leave it just to APIManager itself?

cejanen (changes requested):
Hola,
good idea but I would like to check you my notes to this PRs. Maybe I'm wrong in my ideas as you dig deeper into the problematic part of the code.

matejmolnar (approved):
🙏

cejanen (approved):
🥏

PR #74, opened 2025-03-19
[chore] Update swiftlint

raulbatista94:
This PR updates the siwftlint version to the latest version available. There is a noticeable reduction in build times, especially for SPM modules (like we have on Lea) using SwiftLint as a plugin.


cejanen (approved):
👍

PR #76, opened 2025-06-27
chore: make APIManagerError public

gajddo00:
In order to be able to catch this specific error and map it to something else (for example CancellationError), the type needs to be public.


ipek (approved):

TParizek (approved):
⚠️

PR #77, opened 2026-01-14
Fix: Optional url parameter encoding

matejmolnar:
## Objective of this PR

Solve an issue which broke url parameter encoding in a pretty sneaky manner.

## Example

Let's say you have a router like so

```
private enum ExampleRouter: Requestable {
    case items(limit: Int?)

    var urlParameters: [String: Any]? {
        switch self {
        case let .items(limit):
            [
                "limit": limit
            ]
        }
    }
```

Looks good, but when we run it we find out that the resulting parameter would get encoded as `limit=Optional(10)` instead of `limit=10`. 
Now you might say okay let's fix it by mapping the dictionary values to non optionals like so:

```
            [
                "limit": limit
            ].compactMapValues { $0 }
```

Nice! Everything works now.

After some time there's a new requirement to add another parameter of type String to the request like so:

```
private enum ExampleRouter: Requestable {
    case items(limit: Int?, filter: String?)

    var urlParameters: [String: Any]? {
        switch self {
        case let .items(limit, filter):
            [
                "limit": limit,
                "filter": filter
            ].compactMapValues { $0 }
        }
    }
```

Everything should work right? 

Wrong! The resulting parameters are once again being encoded as `limit=Optional(10)&filter=Optional(test)`. 
What changed?!

Before the dictionary got resolved as `[String: Int?]` by the compiler before applying the `compactMapValues`, however now we have multiples types in the dictionary hence it gets resolved as `[String: Any]` so the `compactMapValues` doesn't do anything.

## Solution

During encoding unwrap the Any type in case it's optional by using mirroring.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: 'iOS 26.1, Xcode 26.2'
-            destination: 'OS=26.1,name=iPhone 16 Pro'
+          - name: 'iOS 26.2, Xcode 26.2'
+            destination: 'OS=26.2,name=iPhone 17 Pro'
             xcode: 'Xcode_26.2'
             runsOn: macos-26
           - name: 'macOS 26, Xcode 26.2'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,14 +23,14 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: 'iOS 18.2, Xcode 16.2'
-            destination: 'OS=18.2,name=iPhone 16 Pro'
-            xcode: 'Xcode_16.2'
-            runsOn: macos-15
-          - name: 'macOS 15, Xcode 16.2'
+          - name: 'iOS 26.1, Xcode 26.2'
+            destination: 'OS=26.1,name=iPhone 16 Pro'
+            xcode: 'Xcode_26.2'
+            runsOn: macos-26
+          - name: 'macOS 26, Xcode 26.2'
             destination: 'platform=macOS'
-            xcode: 'Xcode_16.2'
-            runsOn: macos-15
+            xcode: 'Xcode_26.2'
+            runsOn: macos-26
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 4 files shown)


gajddo00 (dismissed):
🦑

TParizek (dismissed):
Good catch, thanks for fixing the bug!

I see that tests are failing on CI with iOS 18 runtime. Would you mind please bumping test target iOS to iOS 26 to resolve the issue?

cejanen (dismissed):
🚁 

I would like to see running CI, currently we fix it on all projects

RTsta (approved):
🎂

.github/workflows/ci.yml
```diff
@@ -23,14 +23,14 @@ jobs:
...
         include:
-          - name: 'iOS 18.2, Xcode 16.2'
-            destination: 'OS=18.2,name=iPhone 16 Pro'
-            xcode: 'Xcode_16.2'
-            runsOn: macos-15
-          - name: 'macOS 15, Xcode 16.2'
+          - name: 'iOS 26.2, Xcode 26.2'
+            destination: 'OS=26.2,name=iPhone 17 Pro'
+            xcode: 'Xcode_26.2'
+            runsOn: macos-26
```
ipek: Can we please change it so it runs on self-hosted CI? `self-hosted`, `macOS`, `mobile-ci`

.github/workflows/ci.yml
```diff
@@ -23,14 +23,14 @@ jobs:
...
+          - name: 'macOS 26, Xcode 26.2'
             destination: 'platform=macOS'
-            xcode: 'Xcode_16.2'
-            runsOn: macos-15
+            xcode: 'Xcode_26.2'
+            runsOn: macos-26
```
ipek: Can we please change it so it runs on self-hosted CI? `self-hosted`, `macOS`, `mobile-ci`

ipek (approved):

.github/workflows/ci.yml
ipek: Resolving, request moved to a different PR.

.github/workflows/ci.yml
ipek: Resolving, request moved to a different PR.

